PR #1, opened 2013-09-24
Feature/fmn

ralphbean:


fmn/lib/models.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+class User(BASE):
+    __tablename__ = 'users'
+
+    username = sa.Column(sa.String(50), primary_key=True)
+    created_on = sa.Column(sa.DateTime, default=datetime.datetime.utcnow)
+
+    @classmethod
+    def by_username(cls, session, username):
```
lmacken, line 133: This method could be reduced to `return session.query(cls).filter_by(username=username).first()` :neckbeard:

fmn/lib/models.py
```diff
@@ -0,0 +1,215 @@
...
+        pref.delivery_detail = delivery_detail
+        session.add(pref)
+        session.flush()
+        return pref
+
+    @classmethod
+    def load(cls, session, user, context):
+        query = session.query(cls)\
+            .filter_by(user_name=user.username)\
+            .filter_by(context_name=context.name)
```
lmacken, line 202: You can just return `query.first()` and avoid the additional COUNT

PR #2, opened 2013-09-25
Remove unnecessary query.count() calls by using first()

lmacken:
Since query.first() returns None if there are no rows in the table, there is no
need to count the rows first. Discussed in pull #1.

### DIFF
--- a/fmn/lib/models.py
+++ b/fmn/lib/models.py
@@ -90,11 +90,7 @@ class Context(BASE):
 
     @classmethod
     def by_name(cls, session, name):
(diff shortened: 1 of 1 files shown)


dtgay:
+1

ralphbean:
What does `.first()` return if there's no object in the db (i.e., `.count() == 0`)?

ralphbean:
> What does .first() return if there's no object in the db (i.e., .count() == 0)?

It returns `None`, which is great!  http://docs.sqlalchemy.org/en/rel_0_8/orm/query.html#sqlalchemy.orm.query.Query.first

PR #6, opened 2013-11-15
Global variables for all templates and add a first version

pypingou:


ralphbean:
Can you remove the old `template_arguments` func?  Thanks for this!

PR #7, opened 2013-11-20
Feature/pkgdb integration

pypingou:


fmn/lib/pkgdb.py
```diff
@@ -0,0 +1,42 @@
...
+import requests
+log = logging.getLogger(__name__)
+
+## TODO: Move this variable into a configuration file
+PKGDB_API_URL = 'http://209.132.184.188/api/'
+
+## TODO: cache the results of this method
+# This might mean removing the acl and branch argument
+# Can be done using dogpile.cache
+def get_package_of_user(username, acl='commit', branch=None):
```
ralphbean, line 14: nitpick:  can you make this plural like `get_packages_of_user`?

ralphbean:
:+1: aside from the comment above.

We can handle caching later, I think.

PR #8, opened 2013-11-20
Feature/flask openid

pypingou:


pypingou:
I do have to say, I have no idea how this will behave with load-balancing.

ralphbean:
So, do we want to allow people without fas accounts to login and use the fmn
instance that will eventually be hosted by fedora infrastructure?
That seems like a good feature to provide if people want to run their own
fmn instance somewhere.. but I don't know if I want non-fas stuff mixed
in with our infra.

For instance, there's that hypothetical vulnerability with talked about at
Flock where I could go get an openid account with google with the name
'pingou', then log in to fmn and masquerade as you.

pypingou:
On Wed, Nov 20, 2013 at 11:44:25AM -0800, Ralph Bean wrote:

>    So, do we want to allow people without fas accounts to login and use the
>    fmn
>    instance that will eventually be hosted by fedora infrastructure?

That is actually what I had in mind yes, with the idea that upstream or
developers from other communities could then subscribe to watch activity on
their package/tool.

>    That seems like a good feature to provide if people want to run their own
>    fmn instance somewhere.. but I don't know if I want non-fas stuff mixed
>    in with our infra.
> 
>    For instance, there's that hypothetical vulnerability with talked about at
>    Flock where I could go get an openid account with google with the name
>    'pingou', then log in to fmn and masquerade as you.

That is a fair point indeed. Couple of things on this:
- Do we enforce unique username in the DB? (Should we?)
- gmail and yahoo don't provide nickname so we use the fullname information
- We have the confirmation code in place for all our contexts, so someone could
  masquerade as me using my nick (but profiles are private) and could set up
  notifications for me just like I could do it for you w/o creating a dedicated
  email.

But, for sure, we might want to take some time to think it through :)

ralphbean:
> - Do we enforce unique username in the DB? (Should we?)

We do.  https://github.com/fedora-infra/fedmsg-notifications/blob/develop/fmn/lib/models.py#L159

> - gmail and yahoo don't provide nickname so we use the fullname information

Ah, but I could change my fullname to "pingou" in gmail, no?

---

If the _real_ pingou had:
- logged in once before
- setup and confirmed an irc nick
- setup some conservative preferences

I could then login with my google openid as "pingou" and add the "all messages" filter to a chain to spam you... or remove stuff.

---

I think the solution, if we want to keep this, is to extend the `User` model to track the exact openid of the login as well as the username so we can differentiate them.

We do a _lot_ of `if fas.g.user_name == user.username:` conditions in `fmn/web/app.py`.  Those would all need to be changed to compare openid_identifiers instead so we don't confuse http://google.com/openid/pingou with http://pingou.id.fedoraproject.org

abadger:
> I think the solution, if we want to keep this, is to extend the User model to track the exact openid of the login as well as the username so we can differentiate them.
> 
> We do a lot of if fas.g.user_name == user.username: conditions in fmn/web/app.py. Those would all need to be changed to compare openid_identifiers instead so we don't confuse http://google.com/openid/pingou with http://pingou.id.fedoraproject.org

I agree with this but there are some things to consider that might not be immediately obvious:
- You probably want to have an account local to fmn in fmn's schema.
- This account can have multiple openid providers associated with it so that if I was an upstream that used gmail and launchpad I could configure either of those to authenticate my account.
- The fmn-local account probably should have an internal representation for user.username that is more like an id than a username.  We'd store the openid provider's idea of username as a nick name.  that way Fedora users can always have their Fedora username as a nickname  even if someone using google openid with nickname "pingou" created their fmn-local account first.  Just have to remember that nickname would no longer be unique.
- In places where we want to use a short name in the interface (instead of a full name) we'd probably want to identify the user with more than their nickname since it's not unique.  An email address seems like a good compromise between length and uniqueness but people probably don't want their email addresses plastered all over web pages.  Maybe generate some combination of nickname and domain of their openid provider like toshio(id.fedoraproject.org)

pypingou:
On Wed, Nov 20, 2013 at 02:12:19PM -0800, Ralph Bean wrote:

> ```
>    * Do we enforce unique username in the DB? (Should we?)
> ```
> 
>    We do.
>    https://github.com/fedora-infra/fedmsg-notifications/blob/develop/fmn/lib/models.py#L159
> 
> ```
>    * gmail and yahoo don't provide nickname so we use the fullname
>      information
> ```
> 
>    Ah, but I could change my fullname to "pingou" in gmail, no?

That part I am not quite sure as I believe both field 'first name' and 'last
name' have to be filled when creating the gmail account. So you would never be
able to have just 'pingou'.
But there might be a way to define a nick which would then take the priority
over the full name.

> ```
>  ----------------------------------------------------------------------
> ```
> 
>    If the real pingou had:
> 
> ```
>  * logged in once before
>  * setup and confirmed an irc nick
>  * setup some conservative preferences
> ```
> 
>    I could then login with my google openid as "pingou" and add the "all
>    messages" filter to a chain to spam you... or remove stuff.
> 
> ```
>  ----------------------------------------------------------------------
> ```
> 
>    I think the solution, if we want to keep this, is to extend the User model
>    to track the exact openid of the login as well as the username so we can
>    differentiate them.
> 
>    We do a lot of if fas.g.user_name == user.username: conditions in
>    fmn/web/app.py. Those would all need to be changed to compare
>    openid_identifiers instead so we don't confuse
>    http://google.com/openid/pingou with http://pingou.id.fedoraproject.org

If we add the openid_identifier (and I think it would be good to keep the email 
as well) to the user table, then you would not be able to log in as 'pingou' if 
the real one has already logged in. And if he hasn't then he will not be able to
use his nick and will have to use his real name.

The problem is more if someone creates two accounts, one with 'pingou' as full
name and one with 'real name' as full name. Then if 'pingou' from FAS has not 
logged in before, he will never be able to.

pypingou:
On Wed, Nov 20, 2013 at 04:18:07PM -0800, Toshio Kuratomi wrote:

> ```
>  * You probably want to have an account local to fmn in fmn's schema.
>  * This account can have multiple openid providers associated with it so
>    that if I was an upstream that used gmail and launchpad I could
>    configure either of those to authenticate my account.
> ```

I am not sure I see the idea of having 1 account with server openid providers
associated.
Why not just have 2 accounts if upstream wants to get some notifications on one
account and others on another accounts.

But maybe we could also just move the email settings as a property of the chain,
something like: I have a main email set in the User table and for each chain I
create I can either use the main email or set another one (which of course will
have to be confirmed).

> ```
>  * The fmn-local account probably should have an internal representation
>    for user.username that is more like an id than a username. We'd store
>    the openid provider's idea of username as a nick name. that way Fedora
>    users can always have their Fedora username as a nickname even if
>    someone using google openid with nickname "pingou" created their
>    fmn-local account first. Just have to remember that nickname would no
>    longer be unique.
> ```

+1

> ```
>  * In places where we want to use a short name in the interface (instead
>    of a full name) we'd probably want to identify the user with more than
>    their nickname since it's not unique. An email address seems like a
>    good compromise between length and uniqueness but people probably
>    don't want their email addresses plastered all over web pages. Maybe
>    generate some combination of nickname and domain of their openid
>    provider like toshio(id.fedoraproject.org)
> ```

We definitively need to figure out something like this, but the email isn't too
bad as the pages where it would be shown are private anyway.

ralphbean:
> I agree with this but there are some things to consider that might not be immediately obvious:
> - You probably want to have an account local to fmn in fmn's schema.

Cool.  We do have that already.

> - This account can have multiple openid providers associated with it so that if I was an upstream that used gmail and launchpad I could configure either of those to authenticate my account.

At first, this sounds good.  The flow could end up tricky, though -- I seem to remember Lanyrd allowing this.  Here's a simple example scenario:
- I log in with FAS.  A local fmn account is created.
- I click the "also associate my account with this other openid" button.  It works.
- I log out.
- I log in with that other openid, and am correctly associated with my original account.

But here, it can get sticky.  How do we handle this?
- I log in with FAS.  A local fmn account is created.
- I log out.
- I log in with a different openid.  A local fmn account is created for _that_ openid.
- I try to associate the two?  Do we allow trying?  Can we "merge" the two accounts and delete one?  Which one do we delete?

It's possible.. it just gets complicated quickly.

> - The fmn-local account probably should have an internal representation for user.username that is more like an id than a username.  We'd store the openid provider's idea of username as a nick name.  that way Fedora users can always have their Fedora username as a nickname  even if someone using google openid with nickname "pingou" created their fmn-local account first.  Just have to remember that nickname would no longer be unique.

This is going to really hose-up our existing URL scheme, which is based on
nickname firstly.  We'll have to reorganize all of that.

---

I think I am :-1: to this change, since its so potentially disruptive (although, I'm willing to be out-voted :D).  If we could scale it back so that we allow different openid providers, but we disable all but one for production, I would be :+1: to that.  This way, if a different site wanted to deploy fmn they could use their own or another openid.  But, hm.. I guess that is of limited value since lots of other places don't have the nicely-defined openid provider like we do.

Another option would be to punt, and try again after the Dec. 15th due date.

fmn/consumer/backends/irc.py
```diff
@@ -152,7 +152,7 @@ def handle_confirmation_valid_nick(self, nick):
                 acceptance_url=acceptance_url,
                 rejection_url=rejection_url,
                 support_email=self.config['fmn.support_email'],
-                username=confirmation.user_name,
+                username=confirmation.openid,
```
ralphbean, line 155: Should this be `openid=confirmation.openid`?

fmn/consumer/backends/mail.py
```diff
@@ -85,7 +85,7 @@ def handle_confirmation(self, confirmation):
             acceptance_url=acceptance_url,
             rejection_url=rejection_url,
             support_email=self.config['fmn.support_email'],
-            username=confirmation.user_name,
+            username=confirmation.openid,
```
ralphbean, line 88: Should this be `openid=confirmation.openid`?

fmn/consumer/backends/irc.py
pypingou, line 155: We can but then, we'll have to adjust the template as well.

Easy enough if you so desire :)

fmn/web/app.py
```diff
@@ -42,6 +45,44 @@
...
 
 
+@app.before_request
+def check_auth():
+    flask.g.auth = Bunch(
+        logged_in=False,
+        method=None,
+        id=None,
+    )
+    if 'openid' in flask.session:
```
ralphbean, line 68: OK.  Here, it is cool that we can strip off the other URL stuff from yahoo and google so we can have nicer URLs.. but in order to do libravatar, we're going to need to _also_ keep track of the fully-qualified openid.

fmn/web/app.py
```diff
@@ -49,8 +90,11 @@ def shutdown_session(exception=None):
 
 
 def admin(user):
-    return any([team.name in app.config.get('ADMIN_GROUPS', [])
-                for team in user.approved_memberships])
+    if user.endswith('id.fedoraproject.org'):
```
ralphbean: Cool.  :)

fmn/web/app.py
```diff
@@ -49,8 +90,11 @@ def shutdown_session(exception=None):
...
+        username = user.split('.', 1)[0]
```
ralphbean: We could ditch this line...

fmn/web/app.py
```diff
@@ -49,8 +90,11 @@ def shutdown_session(exception=None):
...
+        return user in flask.config.get('FMN_ADMINS', [])
```
ralphbean: ... and here make `FMN_ADMINS` look something like `['ralph.id.fedoraproject.org', 'pingou.id.fedoraproject.org']`.

fmn/web/app.py
```diff
@@ -153,9 +197,11 @@ def about():
 @login_required
 def profile(username):
 
-    if (not flask.g.fas_user or (
-        flask.g.fas_user.username != username and
-            not admin(flask.g.fas_user))):
+    print flask.g.auth.openid, username
```
ralphbean: Spurious print statement here.

fmn/web/app.py
```diff
@@ -153,9 +197,11 @@ def about():
 @login_required
 def profile(username):
```
ralphbean: We should change the language of the argument here from `username` to `openid` (since its not really a username anymore).

fmn/web/app.py
```diff
@@ -182,14 +228,21 @@ def profile(username):
 @app.route('/<not_reserved:username>/<context>/')
 @login_required
 def context(username, context):
```
ralphbean: Same here.  Can we change `username` to `openid`?

fmn/web/app.py
```diff
@@ -182,14 +228,21 @@ def profile(username):
...
     if not context:
         flask.abort(404)
 
-    pref = fmn.lib.models.Preference.get_or_create(SESSION, username, context)
+    try:
+        pref = fmn.lib.models.Preference.get_or_create(
+            SESSION, email=flask.g.auth.email, openid=flask.g.auth.openid,
+            context=context)
+    except sqlalchemy.exc.IntegrityError:
+        flask.abort(403, "Someone has already loged in with the email: %s" %
```
ralphbean: Ah, here is where unique email is used.

fmn/web/app.py
pypingou: Then we can ditch both and just have a list of 'username.id.fp.o' in the config file

fmn/web/app.py
pypingou: gmta :)

fmn/web/app.py
```diff
@@ -385,15 +446,15 @@ def handle_filter():
...
         raise APIError(403, dict(reason="%r is not %r" % (
-            flask.g.fas_user.username, username
+            flask.g.auth.openid, username
         )))
 
     if method not in ['POST', 'DELETE']:
         raise APIError(405, dict(reason="Only POST and DELETE accepted"))
 
-    user = fmn.lib.models.User.by_username(SESSION, username)
+    user = fmn.lib.models.User.by_openid_provider(SESSION, username)
```
ralphbean: Here, like the other places, let's change `username` to `openid`.

fmn/consumer/backends/irc.py
ralphbean, line 155: Yeah, let's clean it all the way through.  `username` gets the _kibosh_!

fmn/web/app.py
ralphbean: Yeah, let's do.  :)

pypingou:
```
Ran 11 tests in 2.224s

OK
```

pypingou:
:+1: for me

ralphbean:
:+1: from me!

PR #11, opened 2013-11-22
Update license header to LGPLv2+

### LINKED ISSUE #5: Clear license
The two setup-*.py files are saying the license is LGPLv2+ while the header in https://github.com/fedora-infra/fedmsg-notifications/blob/feature/webapp/fmn/lib/models.py seems to say GPLv2+

pypingou:
Fixes https://github.com/fedora-infra/fmn/issues/5

### DIFF
--- a/fmn/lib/models.py
+++ b/fmn/lib/models.py
@@ -4,14 +4,15 @@
 #
 # This copyrighted material is made available to anyone wishing to use,
 # modify, copy, or redistribute it subject to the terms and conditions
-# of the GNU General Public License v.2, or (at your option) any later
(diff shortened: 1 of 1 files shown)


pypingou:
Two things: I did not precise which of the version 2.0 or 2.1 we use and I did not add a link to http://www.gnu.org/licenses/old-licenses/lgpl-2.0.txt (or http://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt ) which we could include in the sources.

ralphbean:
Cool.  Thanks, @pypingou.

PR #136, opened 2016-10-10
Update the Ansible role to not update git repos

jeremycline:
The Ansible role Vagrant uses conveniently checks out the various git
repositories for you. However, on subsequent provisions, it tries to
update the repositories which fails if you don't have an remote called
`origin`, or have uncommitted changes, etc.

This will cause Ansible to initially check out the repositories, but
leave them alone afterwards.

Signed-off-by: Jeremy Cline jeremy@jcline.org


pypingou:
:+1: for me

PR #139, opened 2016-12-05
Add a Sphinx documentation project

jeremycline:


database.txt
```diff
@@ -1,38 +0,0 @@
- +-------------+      +-----------------+   +---------------+   +------------+
- | users       |      | preferences     |   | filters       |   | rules      |
```
pypingou, line 2: Don't want to keep this file?
It's what was used to generate the image and thus can be handy if there is a need to update it

pypingou:
It seems to me that quite a bit of doc is lost in translation here :(

database.txt
jeremycline: Was it made by hand?

I was thinking it would be easiest to build the database diagram from the database itself or the SQLAlchemy models. I found [this](https://bitbucket.org/zzzeek/sqlalchemy/wiki/UsageRecipes/SchemaDisplay), which produces:
dbschema

It's not great, admittedly, but it's automated and will always be up-to-date. Maybe there are better tools out there, as well, but that one was super simple to use and would be easy to add to the docs Makefile.

pypingou: It was likely made by hand using asciiflow I suspect.

Nice tool, less nice looking but can be handy in more cases

jeremycline:
I updated the docs makefile to automatically build the schema diagram from the latest models in fmn.lib. I also pulled a more accurate system diagram from fmn.consumer to replace the one I removed. The rest of the documentation I removed looks more like old planning documents to me. Are there particular pieces you think we should save?

README.rst
```diff
@@ -1,187 +1,32 @@
...
-          |       +-> does not pertain to
-          |           package 'nethack'
-          |
-          +--->Filter2
-                  |
-                  +-> is a bodhi message
-                  |
-                  +-> pertains to a package
-                      owned by 'lmacken'
-
```
pypingou, line 158: Is this a diagram we want to keep?

pypingou:
Maybe one more diagram, otherwise 👍 for me.

We may want to publish this doc somewhere, having it outside the README makes it potentially harder to find/read, so maybe we could build it somewhere and link to it in the README

jeremycline:
Okay, I added the rules diagram and autodoc'd the rules package. I also built it for gh-pages so you can see it at https://fedora-infra.github.io/fmn/index.html and that's linked in README.rst

pypingou:
Nitpicking:
- Maybe we could put the glossary above the list of rules, as otherwise it's easy to miss
- I'm not finding the link to the reference page --- or could it be that the github page isn't up to date?

Otherwise 👍 for me

jeremycline:
Good nitpicks :smile:. I split the reference out from the table of content at the top and put it at the bottom, which looks much less cluttered. I also updated the github page.

Although the GitHub pages are quick to get going, I'm tempted to put the documentation on Read the Docs because it'll autobuild the docs and keep versions of documentation around. Any objections to that?

pypingou:
> Although the GitHub pages are quick to get going, I'm tempted to put the documentation on Read the Docs because it'll autobuild the docs and keep versions of documentation around. Any objections to that?

Quite the opposite, RTD is built on FOSS softwares and tools :)

PR #140, opened 2016-12-12
Merge fmn.lib, fmn.consumer, and fmn.rules into fmn

jeremycline:
As discussed in https://github.com/fedora-infra/fmn.lib/pull/66, this merges fmn.lib, fmn.consumer, and fmn.rules into a single repository. History is preserved here, although I did trim the changelog.

I'm still playing with the development environment for this so it may need a bit of additional polishing, but it's pretty close. The one big thing left to do is decide how to version this. Does anyone have an opinion on that front?


pypingou:
I don't know if this is worth reviewing, you investigated the merge and checked that we do not loose history, so 👍 for me, but I'm not going to review the 250 commits and 119 files changed :)

jeremycline:
😄 That's fair

PR #142, opened 2016-12-19
Make fedmsg[crypto] optional

jeremycline:
Intially I set the requirement for fedmsg to include all its extra
requirements. However, you do not need the crypto requirements to
install or run FMN, only to validate the signatures on messages. This
makes installing fmn simpler on platforms like Travis or RTD since they
may not have SWIG which is required to install M2crypto from PyPI.

It also adds the requirements to build the docs into the requirements file
that held the test dependencies.

### DIFF
--- a/ansible/roles/fmn-dev/tasks/main.yml
+++ b/ansible/roles/fmn-dev/tasks/main.yml
@@ -79,32 +79,40 @@
   args:
     chdir: /home/{{ ansible_env.SUDO_USER }}/devel
 
-- name: Install Python 2 tests requirements
+- name: Install Python 2 dev requirements for fmn
+  become_user: "{{ ansible_env.SUDO_USER }}"
+  pip:
+    requirements: "dev-requirements.txt"
+    virtualenv: /home/{{ ansible_env.SUDO_USER }}/.virtualenvs/python2-fmn/
+    virtualenv_python: python2
+  args:
+    chdir: /home/{{ ansible_env.SUDO_USER }}/devel
+
+
+- name: Install Python 2 dev requirements for fmn.web and fmn.sse
   become_user: "{{ ansible_env.SUDO_USER }}"
   pip:
     requirements: "{{ item }}/tests-requirements.txt"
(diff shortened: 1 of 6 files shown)


ansible/roles/fmn-dev/tasks/main.yml
```diff
@@ -79,32 +79,40 @@
...
 
 # NOTE: Sadly fmn.web isn't Python 3 compatible yet
-- name: Install Python 3 tests requirements
+- name: Install Python 3 dev requirements for fmn
   become_user: "{{ ansible_env.SUDO_USER }}"
-  shell: "source ~/.bashrc && workon python3-fmn && pip install -r {{ item }}/tests-requirements.txt"
-  with_items:
-    - "."
-    - "fmn.sse"
+  shell: "source ~/.bashrc && workon python3-fmn && pip install -r dev-requirements.txt"
```
bowlofeggs, line 107: Why not use the pip module like you did above?

ansible/roles/fmn-dev/tasks/main.yml
```diff
@@ -79,32 +79,40 @@
...
   args:
     chdir: /home/{{ ansible_env.SUDO_USER }}/devel
 
-- name: Install packages necessary to build the documentation
+- name: Install Python 3 dev requirements for fmn.sse
   become_user: "{{ ansible_env.SUDO_USER }}"
-  shell: "source ~/.bashrc && workon python3-fmn && pip install sphinx sqlalchemy_schemadisplay"
+  shell: "source ~/.bashrc && workon python3-fmn && pip install -r tests-requirements.txt"
```
bowlofeggs, line 113: Same here?

bowlofeggs (approved):
LGTM!

ansible/roles/fmn-dev/tasks/main.yml
jeremycline: It doesn't work with Python 3 😞

PR #150, opened 2016-12-19
Use the explicit relative import syntax for backend imports

jeremycline:
Implicit relative imports are both confusing and incompatible with
Python 3.

See https://www.python.org/dev/peps/pep-0328/


pypingou:
👍

PR #152, opened 2016-12-19
Change all print statements to calls to print

jeremycline:

### DIFF
--- a/alembic/versions/1b6e419ea1a6_add_threshold1_to_existing_prefs.py
+++ b/alembic/versions/1b6e419ea1a6_add_threshold1_to_existing_prefs.py
@@ -5,6 +5,7 @@ Revises: 24533471e302
 Create Date: 2015-05-05 10:11:55.738516
 
 """
+from __future__ import print_function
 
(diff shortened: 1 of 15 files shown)


alembic/versions/1b6e419ea1a6_add_threshold1_to_existing_prefs.py
```diff
@@ -39,10 +39,10 @@ def upgrade():
     valid_paths = fmn.lib.load_rules(root='fmn.rules')
 
     filters = session.query(fmn.lib.models.Filter).filter_by(name=target).all()
-    print "Found %i filters" % len(filters)
+    print("Found %i filters" % len(filters))
```
pypingou, line 43: Don't we want to import from __future__ then?

jeremycline: I used 2to3 and it very unkindly does not add from __future__ imports. I've fixed that.

pypingou:
👍 for me

PR #154, opened 2017-01-03
Modify the fmn-web systemd unit to use the flask app runner

jeremycline:
This changes the developer systemd unit to make use of the built-in
Flask application runner (flask >= 0.11 required).


pypingou:
👍

PR #155, opened 2017-01-05
Remove the Github Pages Sphinx extension

jeremycline:
readthedocs.org has an older version of Sphinx that doesn't include this
extension by default, and we are not going to use it, so this turns it
off.

This is the last thing necessary to get RTD working for FMN. I built
[this branch](http://jcline-fmn.readthedocs.io/en/bye-ghpages/) to make sure.


pypingou:
👍 for me

PR #158, opened 2017-01-05
Add requirements file to manifest

jeremycline:
Since the requirements files are read in the setup.py, they need to be
accessible for an sdist to be installed. This includes them in the
manifest.


pypingou:
👍 for me

PR #159, opened 2017-01-05
Remove the empty line from dev-requirements.txt

jeremycline:
The little parsing function in setup.py doesn't handle newlines very
well, and distutil gets unhappy if a requirement is an empty string.
This removes the empty line.

PS: I'm really sorry for the flurry of tiny PRs


pypingou:
Easy enough :) 👍

PR #169, opened 2017-02-18
requeue

puiterwijk:
Signed-off-by: Patrick Uiterwijk <puiterwijk@redhat.com>

### DIFF
--- a/fmn/consumer/backend.py
+++ b/fmn/consumer/backend.py
@@ -1,5 +1,11 @@
-# FMN worker figuring out for a fedmsg message the list of recipient and
-# contexts
+# -*- coding: utf-8 -*-
+"""
+The FMN backend is a `Twisted`_ application that consumes messages that have
+been queued by FMN workers and sends them to the user over the configured
+medium (IRC, email, etc) by using "backends".
+
+.. _Twisted: https://twistedmatrix.com/
+"""
 from __future__ import print_function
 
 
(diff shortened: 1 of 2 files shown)


fmn/consumer/backend.py
```diff
@@ -152,22 +152,33 @@ def read(queue_object):
...
+                backend.handle_batch(session,
+                                     data['recipient'],
+                                     data['queued_messages'])
+                ch.basic_ack(delivery_tag=method.delivery_tag)
+            else:
+                print("Unknown function")
+            return
+        except Exception:
+            logging.error('Message failed, requeueing', exc_info=True)
+            ch.basic_publish(exchange='',
```
jeremycline: In your original PR, you used ``basic_nack`` which should requeue the message (by default).  Is there a reason you switched to publishing it again?

puiterwijk: As discussed on IRC: I am now manually requeueing to put it back at the end of the queue.
The basic_nack puts it back at the front of the queue, still leaving the worker unable to send out any messages until this message is fixed.

I'm going to say no to dropping messages entirely in general.

We should note that this PR makes it so that messages might be sent out in a different order then received, but to a certain extent that'll always be the case with multiple workers.

jeremycline: I think this will work fine for now, but in the future I'd like to have a better method of dealing with undelivered messages. This will result in a message that gets re-queued for eternity/when a bugfix is applied.

Better retry logic and graceful handling of missing external services in combination with, say, a [dead letter exchange](https://www.rabbitmq.com/dlx.html) to record the offending messages might be good. Then maybe a service that takes dead letters and sends a generic alert to users with a datanommer link? Not sure if that's feasible at all, but it's food for thought.

PR #171, opened 2017-02-23
Document the FMN consumer package

jeremycline:
This adds docblocks to the various FMN consumer pieces.

### DIFF
--- a/fmn/consumer/__init__.py
+++ b/fmn/consumer/__init__.py
@@ -1,3 +1,23 @@
+"""
+The FMN consumer package contains all the service that process the incoming
+`fedmsg`_ messages. The messages are processed in three steps:
+
+1. The `fedmsg consumer`_ defined in :mod:`fmn.consumer.consumer` subscribes
+   to every fedmsg topic (``*``). It places these messages into a RabbitMQ
+   message queue called ``worker``. This message broker provides message
+   durability for FMN as it processes the messages.
+
+2. One or more :mod:`fmn.consumer.worker` processes are started and consume
+   messages from the ``worker`` message queue. These worker processes take
+   each message and determine who should receive notifications based on their
+   message filters. It then records this information in the message and
+   publishes the message to the ``backend`` queue.
+
+3. A single :mod:`fmn.consumer.backend` process is run and is responsible for
+   sending the messages to users via IRC, email, etc. It consumes the messages
+   from the ``backend`` queue and dispatches them. It defines a backend interface
+   and new backends can be added to allow for new message mediums.
+"""
 from .consumer import FMNConsumer  # noqa
 from .producer import ConfirmationProducer  # noqa
 from .producer import DigestProducer  # noqa
--- a/fmn/consumer/backend.py
+++ b/fmn/consumer/backend.py
@@ -91,6 +91,12 @@ for key in CONFIG['fmn.backends']:
 
 
 def get_preferences():
+    """
+    Load all preferences from the FMN database and return them.
+
+    Returns:
+        dict: A big dictionary of user preferences.
+    """
     print('get_preferences')
     prefs = {}
     for p in session.query(fmn.lib.models.Preference).all():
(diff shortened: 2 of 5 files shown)


fmn/consumer/__init__.py
```diff
@@ -1,3 +1,23 @@
...
+   message queue called ``worker``. This message broker provides message
+   durability for FMN as it processes the messages.
+
+2. One or more :mod:`fmn.consumer.worker` processes are started and consume
+   messages from the ``worker`` message queue. These worker processes take
+   each message and determine who should receive notifications based on their
+   message filters. It then records this information in the message and
+   publishes the message to the ``backend`` queue.
+
+3. A single :mod:`fmn.consumer.backend` process is run and is responsible for
```
pypingou, line 16: In theory this could be "one or more" as well I believe

fmn/consumer/backend.py
```diff
@@ -103,6 +109,16 @@ def get_preferences():
 
 
 def update_preferences(openid, prefs):
+    """
+    Update an existing preference dictionary loaded by :func:`get_preferences`.
```
pypingou: It updates the prefs but for a certain user/profile not the entire dict :)

fmn/consumer/consumer.py
```diff
@@ -80,9 +121,23 @@ def __init__(self, *args, **kwargs):
...
+        Returns:
+            sqlalchemy.orm.scoping.scoped_session: An SQLAlchemy scoped session.
+                Calling it returns the current Session, creating it using the
+                scoped_session.session_factory if not present.
+        """
         return fmn.lib.models.init(self.uri)
 
     def consume(self, raw_msg):
+        """
+        This method is called when a message arrives.
```
pypingou: arrives on the bus?

fmn/consumer/consumer.py
```diff
@@ -92,6 +147,13 @@ def consume(self, raw_msg):
             raise
 
     def work(self, session, raw_msg):
+        """
+        This method is called when a message arrives.
```
pypingou: arrives from?

fmn/consumer/consumer.py
```diff
@@ -1,4 +1,12 @@
-# An example fedmsg koji consumer
+"""
+This is a `fedmsg consumer`_ that subscribes to every topic on the message bus
+it is connected to. It has two tasks. The first is to place all incoming
+messages into a RabbitMQ message queue. The second is to manage the FMN caches.
+
+FMN makes heavy use of caches since it needs to know who owns what packages and
+what user notification preferences are, both of which require expensive API
+queries to `FAS`_, `pkgdb`_, or the database.
```
pypingou, line 8: The links in this paragraph do not seem to be present, is it ok ?

fmn/consumer/worker.py
```diff
@@ -54,6 +71,16 @@ def get_preferences():
 
 
 def update_preferences(openid, prefs):
+    """
+    Update an existing preference dictionary loaded by :func:`get_preferences`.
```
pypingou: I'd have the same comment as above here, I'd specify we're updating just a part of the dict :)

pypingou:
A few comments but overall it reads nicely, it's cool :)

fmn/consumer/__init__.py
jeremycline: Isn't this limited to one at the moment because we have to have one connection to IRC, throttling, etc?

pypingou: Isn't twisted supposed to allow running more than one?

I know i've wanted to use only one for simplicity reasons but it was my understanding that we could scale (which is also a reason why the service is written as a template not a single service)

fmn/consumer/consumer.py
jeremycline: I told myself I'd do it all at the end and I forgot :(

fmn/consumer/__init__.py
jeremycline: I think at the moment there's a lot of assumptions in the code that it's only running as one instance. It'll always load the same IRC nickname for every process you start, for example. Also, I'm not 100% sure, but because the backend is also queuing messages in the database in a 'queued_messages' table it _might_ not work well with multiple backends.

pypingou:
👍 for me

PR #172, opened 2017-02-24
Start creating a scoped session object in the models module

jeremycline:
Since we load the configuration at module import time anyway, there's no
reason we can't create the database engine at import time as well. The
engine only connects to the database when connect is explicitly called
or something that requires a connection is called[0]. Furthermore, we
can also create the scoped session object.

This adds the ability to create the database to the fmn-createdb script
and stops that from calling ``fmn.lib.models.init``. That function has
been deprecated.

[0] http://docs.sqlalchemy.org/en/latest/core/engines.html#sqlalchemy.create_engine

### DIFF
--- a/fmn/lib/db.py
+++ b/fmn/lib/db.py
@@ -20,50 +20,68 @@ This module provides a command to configure a database for FMN.
 """
 from __future__ import print_function, unicode_literals
 
-import sys
+import argparse
 
-import fedmsg.config
-import fmn.lib.models
+from fmn.lib import models
+
 
 def main():
     """
     The entry point for the database commands.
     """
-    config = fedmsg.config.load_config()
-    uri = config.get('fmn.sqlalchemy.uri')
-    if not uri:
-        raise ValueError("fmn.sqlalchemy.uri must be present")
+    parser = argparse.ArgumentParser(description='FMN database manager')
+    parser.add_argument(
+        '--create',
+        '-c',
+        dest='create',
+        action='store_true',
+        help='Create the database tables'
+    )
+    parser.add_argument(
+        '--with-dev-data',
+        '-d',
+        dest='dev',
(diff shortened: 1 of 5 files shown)


fmn/lib/db.py
```diff
@@ -20,50 +20,68 @@
 """
 from __future__ import print_function, unicode_literals
 
-import sys
+import argparse
```
bowlofeggs, line 23: I would recommend [click](http://click.pocoo.org/5/) over argparse in general.

fmn/lib/models.py
```diff
@@ -70,28 +81,25 @@ def init(db_url, alembic_ini=None, debug=False, create=False):
...
+            information with regards to the database engine, the host to
+            connect to, the user and password and the database name.
+            ie: <engine>://<user>:<password>@<host>/<dbname>
+        alembic_ini (str): path to the alembic ini file. This is necessary
+            to be able to use alembic correctly, but not for the unit-tests.
+        debug (bool): a boolean specifying wether we should have the verbose
+            output of sqlalchemy or not.
+
+    Returns:
+        scopedsession: An SQLAlchemy scoped session.
```
bowlofeggs, line 98: This docblock change seems unrelated to the purpose of the rest of this commit. I recommend splitting it into a separate commit.

bowlofeggs (approved):
Just a couple suggestions at your option. LGTM!

fmn/lib/db.py
pypingou: one is in stdlib and the other adds a new dependency, so there is a tradeoff :)

jeremycline: Yeah, click looks good, but for this (which is only meant to create the database) isn't going to be a complex CLI. If it starts to grow, I'll reconsider using click!

PR #175, opened 2017-03-15
Unify the various preference loading/updating functions

jeremycline:
This is made up of two commits.

The first commit adds a query property to the base SQLAlchemy model.
This allows for the ``ModelClass.query(...)`` pattern rather than
passing the session object around. The query property uses the scoped
session.

The second commit (which relies on the first) unifies the various preference
loading/updating functions Previously there were _three_ different functions
to load preferences and _two_ functions update them. The also used different
(undocumented) formats. This gets rid of two of those loading functions and 
moves the update function into the same module as the last loading function.

### DIFF
--- a/fmn/consumer/backend.py
+++ b/fmn/consumer/backend.py
@@ -14,21 +14,19 @@ import logging
 import smtplib
 import time
 
+from fedmsg_meta_fedora_infrastructure import fasshim
+from pika.adapters import twisted_connection
+from twisted.internet import defer, reactor, protocol, task
 import pika
 import fedmsg
 import fedmsg.meta
 import fedmsg_meta_fedora_infrastructure
 
-from pika.adapters import twisted_connection
-from twisted.internet import defer, reactor, protocol, task
-
 import fmn.lib
 import fmn.rules.utils
 import fmn.consumer.backends as fmn_backends
 import fmn.consumer.producer as fmn_producers
-
 import fmn.consumer.fmn_fasshim
-from fedmsg_meta_fedora_infrastructure import fasshim
 
 
 logging.basicConfig(level=logging.DEBUG)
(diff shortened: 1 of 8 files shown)


fmn/lib/models.py
```diff
@@ -59,6 +59,16 @@
...
 class FMNBase(object):
+    """
+    Base class for the SQLAlchemy model base class.
+
+    Attributes:
+        query (sqlalchemy.orm.query.Query): a class property which produces a
+            Query object against the class and the current Session when called.
+    """
+
+    query = Session.query_property()
```
bowlofeggs, line 70: I like this.

fmn/consumer/backend.py
```diff
@@ -14,21 +14,19 @@
...
 
+from fedmsg_meta_fedora_infrastructure import fasshim
+from pika.adapters import twisted_connection
+from twisted.internet import defer, reactor, protocol, task
 import pika
 import fedmsg
 import fedmsg.meta
 import fedmsg_meta_fedora_infrastructure
 
-from pika.adapters import twisted_connection
```
bowlofeggs, line 22: This seems like a PEP-8 move - I recommend doing this in a separate commit to keep this one cleaner and focused.

fmn/consumer/worker.py
```diff
@@ -20,14 +20,13 @@
...
-
-import fmn.lib
-import fmn.rules.utils
 import fedmsg
 import fedmsg.meta
 import fedmsg_meta_fedora_infrastructure
 import pika
 
+import fmn.lib
+import fmn.rules.utils
```
bowlofeggs, line 29: I recommend making these changes in a separate PEP-8 commit as well.

fmn/lib/__init__.py
```diff
@@ -1,19 +1,19 @@
 """ Fedora Notifications internal API """
 
-import fmn.lib.hinting
-import fmn.lib.models
-
+from collections import defaultdict
```
bowlofeggs, line 3: Same here for PEP-8.

fmn/tests/lib/test_lib.py
```diff
@@ -1,65 +1,144 @@
...
+class LoadPreferenceTests(BaseTestCase):
 
-
-class TestRecipients(fmn.tests.Base):
     def setUp(self):
-        super(TestRecipients, self).setUp()
+        super(LoadPreferenceTests, self).setUp()
+        user = models.User(openid='jcline', openid_url='http://jcline.id.fedoraproject.org/')
+        user2 = models.User(
+            openid='bowlofeggs', openid_url='http://bowlofeggs.id.fedoraproject.org/')
```
bowlofeggs, line 38: Oh yeah, I'm user2!

bowlofeggs (approved):
LGTM - the PEP-8 suggestion is at your option, and I don't feel a need to re-review if you decide to make that change.

PR #183, opened 2017-04-26
Adjust the way preferences are accessed in the backend

jeremycline:
Previously it had access to the model directly, but when the preference
loading got unified it now has a dictionary.

### DIFF
--- a/fmn/consumer/backend.py
+++ b/fmn/consumer/backend.py
@@ -206,11 +206,11 @@ def read(queue_object):
     for recipient in recipients:
         user = recipient['user']
         t = time.time()
-        pref = PREFS.get('%s__%s' % (user, context))
+        pref = PREFS.get('%s_%s' % (user, context))
         print("pref retrieved in: %0.2fs" % (time.time() - t))
 
         try:
-            if not pref.should_batch:
+            if pref.get('batch_delta') is None and pref.get('batch_count') is None:
                 print("    Calling backend %r with %r" % (backend, recipient))
                 t = time.time()
                 backend.handle(session, recipient, raw_msg)


fmn/consumer/backend.py
```diff
@@ -206,11 +206,11 @@ def read(queue_object):
     for recipient in recipients:
         user = recipient['user']
         t = time.time()
-        pref = PREFS.get('%s__%s' % (user, context))
+        pref = PREFS.get('%s_%s' % (user, context))
         print("pref retrieved in: %0.2fs" % (time.time() - t))
 
         try:
-            if not pref.should_batch:
+            if pref['batch_delta'] is None and pref['batch_count'] is None:
```
pypingou: Should we use .get() here?

fmn/consumer/backend.py
```diff
@@ -206,11 +206,11 @@ def read(queue_object):
     for recipient in recipients:
         user = recipient['user']
         t = time.time()
-        pref = PREFS.get('%s__%s' % (user, context))
+        pref = PREFS.get('%s_%s' % (user, context))
```
pypingou, line 209: Did we fix the place where this is set?

jeremycline: This was a silly thing I managed to somehow typo (obviously there's no test coverage) when converting stuff. The dictionary uses the key format fixed here.

fmn/consumer/backend.py
jeremycline: I don't think it should ever be missing, but it wouldn't hurt. Good call.

pypingou:
👍 for me

PR #184, opened 2017-04-26
Adjust the test runners to handle the trial tests

jeremycline:
Several tests for the FMN SSE server use the Twisted trial test
framework, which the native test runner does not handle. pytest handles
them just fine, though.

Since we're not distributing fmn as separate eggs anymore, we don't need
the namespace package and as it stood it made py.test fail in tox.

This also switched TravisCI to run the tests via Tox rather than
switching it to using py.test so the tests are run uniformly locally and
in CI.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,13 @@
+[run]
+branch = True
+include =
+    fmn/*
+
+[report]
+precision = 2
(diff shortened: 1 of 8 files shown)


pypingou:
Looks fine to me

PR #190, opened 2017-05-16
Cache a mapping of rule strings to rule objects

### LINKED ISSUE #189: Rework the preference loading code to increase performance
At the moment, it takes about an hour (!) to [load all the preferences](https://github.com/fedora-infra/fmn/blob/develop/fmn/lib/__init__.py#L132) from the Fedora deployment (~8000 preferences).

One potential easy fix is problem to switch from the default lazy loading approach ('select') to an eager loading via ['joined'](http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html#joined-eager-loading) so all relationships are loaded in a single query. Part of the long load time is probably due to the high number of queries combined with latency. 

edit:  I profiled preference loading and it turns out it's spending 99% (literally) of its time calling __import__ on strings from the database that point to rule classes.

jeremycline:
The preference loading resulted in loading rule objects via
``__import__`` for each rule in the database. For 170 users with default
preferences, this is approximately 30,000 rows and therefore 30,000
imports. Using a dictionary cache decreased the load time by nearly an
order of magnitude (in decimal) from 56 seconds to ~6 seconds. With
larger rulesets this should increase by even more than an order of
magnitude as there will be a larger percentage of cache hits.

This is a memory-backed cache that never prunes its entries. However,
there are currently only about 300 rules in FMN and that is not a
particularly large dictionary.

fixes #189

### DIFF
--- a/fmn/tests/lib/test_models.py
+++ b/fmn/tests/lib/test_models.py
@@ -23,6 +23,7 @@ import unittest
 from sqlalchemy.orm.query import Query
 import mock
 
+from fmn.rules import user_package_filter
 import fmn.lib.models
 import fmn.tests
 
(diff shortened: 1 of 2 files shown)


ralphbean (approved):
:rocket:

fmn/tests/lib/test_models.py
```diff
@@ -223,3 +223,34 @@ def test_queued_message_dequeue(self):
...
+    def test_results_same_object(self):
+        """Assert that the exact same object is returned from multiple calls."""
+        r1 = fmn.lib.models._cached_load_class('fmn.rules:user_package_filter')
+        r2 = fmn.lib.models._cached_load_class('fmn.rules:user_package_filter')
+
+        self.assertTrue(r1 is r2)
+
+    def test_results_expected_object(self):
+        """Assert the object loaded is what we expect it to be."""
+        from fmn.rules import user_package_filter as expected_rule
```
bowlofeggs: Why the mid-method import?

bowlofeggs (approved):

fmn/tests/lib/test_models.py
jeremycline: I thought it made the test clearer. I felt it was worth committing a small crime for the sake of clarity, but maybe it's just surprising to see a mid-method import.

PR #197, opened 2017-05-26
taskotron rules: support wildcard matching

kparal:
Particular tasks can now be matched using wildcards.

Fixes: https://phab.qa.fedoraproject.org/T953

### DIFF
--- a/fmn/rules/taskotron.py
+++ b/fmn/rules/taskotron.py
@@ -1,4 +1,13 @@
 from fmn.lib.hinting import hint, prefixed as _
+import fnmatch
+
+
+RELEASE_CRITICAL_TASKS = [
+    # if you update this, don't forget to also update the docstring for
+    # taskotron_release_critical_task()
+    'dist.depcheck',
+    'dist.upgradepath',
+]
 
 
 @hint(topics=[_('taskotron.result.new')])
@@ -8,7 +17,7 @@ def taskotron_result_new(config, message, **kwargs):
     This rule lets through messages from the `taskotron
     <https://taskotron.fedoraproject.org>`_ about new task result.
     """
-    return message['topic'].endswith('taskotron.result.new')
+    return message['topic'].endswith('.taskotron.result.new')
 
 
 @hint(categories=['taskotron'], invertible=False)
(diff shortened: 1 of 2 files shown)


fmn/rules/taskotron.py
```diff
@@ -29,8 +38,14 @@ def taskotron_task(config, message, task=None):
     if not task:
         return False
 
-    tasks = [item.strip().lower() for item in task.split(',')]
-    return message['msg']['task'].get('name').lower() in tasks
+    name = message['msg']['task'].get('name')
```
tflink: Is there a reason that you got rid of `.lower()` here? That seems like something that could break existing rules if folks are using any upppercase letters in their settings

tflink (commented):
Other than the .lower() issue, looks good to me

fmn/rules/taskotron.py
kparal: Because testcase names are not case-insensitive, afaik. I'd say it's better to change now than later. But if you think it's a minor issue which is likely not going to occur in the future, no problem, I'll put it back.

kparal:
Updated.

bowlofeggs:
I recommend adding tests for this change.

kparal:
Updated. There are no other tests for any other rule definitions, so I copied the skeleton from `rules/test_generic.py` and adjusted a little. The tests pass on my machine.

fmn/rules/taskotron.py
```diff
@@ -29,8 +46,15 @@ def taskotron_task(config, message, task=None):
     if not task:
         return False
 
+    name = message['msg']['task'].get('name').lower()
     tasks = [item.strip().lower() for item in task.split(',')]
-    return message['msg']['task'].get('name').lower() in tasks
+    tasks = [item for item in tasks if item]
```
jeremycline: Maybe drop this line and just do ``if task and fnmatch.fnmatchcase(name, task):``

jeremycline (changes requested):
Thanks @kparal, especially for starting on the rules tests (I need to devote some time to building a solid test suite for the rules)!

This looks good to me - I had one small note. CI is passing and coverage looks perfect so after that note is addressed I'm happy to merge this.

fmn/rules/taskotron.py
kparal: I guess I prefer it the other way, but sure, why not.

kparal:
Updated. Please note that I'll have some further minor tweaks to taskotron rules once Beta freeze is lifted. I'm mentioning it just in case you're planning a new fmn release.

jeremycline (approved):

jeremycline:
Thanks for humoring me, I know it was a pretty nit-picky note :smile:

I was planning on making a new fmn release in the next day or two so it can go into staging. Then I can push it out soon after the beta freeze is over. If you want to submit those changes now I'm happy to get them into that release.

PR #198, opened 2017-05-31
update critical tasks in taskotron rules

kparal:
Hello, please also include this change in the upcoming FMN release.

### DIFF
--- a/fmn/rules/taskotron.py
+++ b/fmn/rules/taskotron.py
@@ -5,7 +5,7 @@ import fnmatch
 RELEASE_CRITICAL_TASKS = [
     # if you update this, don't forget to also update the docstring for
     # taskotron_release_critical_task()
-    'dist.depcheck',
+    'dist.rpmdeplint',
     'dist.upgradepath',
 ]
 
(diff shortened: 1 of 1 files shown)


fmn/rules/taskotron.py
```diff
@@ -5,7 +5,8 @@
 RELEASE_CRITICAL_TASKS = [
```
pypingou, line 5: Suddenly I wonder if it would make sense to have these in a configuration file :)

pypingou:
One comment, otherwise looks good to me

fmn/rules/taskotron.py
kparal: That's up to the fmn devs, I guess. If they prepare some plumbing, we can certainly adjust the code.

However, I'd like to push this change now, and deal with config files later.

jeremycline: It does make sense to me for this to be in a configuration file, but currently the configuration just using fedmsg's configuration and needs a lot of refactoring anyway. I'll file an issue about it.

jeremycline (approved):
I filed #199 for the configuration issue, but otherwise this looks fine. Thanks!

kparal:
Thanks! That's all we needed wrt taskotron changes before a new release.

kparal:
I looked at staging and the updated rule documentation doesn't render properly. I wonder why?

fmn

pypingou:
Could it be because of the ``::`` which normally indicate a code-block?

kparal:
Thanks @pypingou . Already resolved in #203.

PR #201, opened 2017-06-01
Backport the kwarg_function_key_generator function

jeremycline:
This backports a single function added in dogpile.cache-0.6.2. We
attempt to import it directly from dogpile.cache and fall back to this
backported version if it's not available.

In addition, I added a test environment so this sort of thing is less likely to slip through in the future.


fmn/dogpile_backports.py
```diff
@@ -0,0 +1,84 @@
...
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
+# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+# SUCH DAMAGE.
```
bowlofeggs, line 27: Does your setup.py have a license expressed in it? If so, I recommend adding this license there. Also, I recommend mentioning that this new license is present in the release notes.

bowlofeggs (approved):

fmn/dogpile_backports.py
jeremycline: I've made it so

bowlofeggs: Cool, my LGTM still stands.

PR #205, opened 2017-07-12
Get the fmn.rules pkgdb tests working again.

ralphbean:
It appears that Jeremy's package ownerships changed.


ralphbean:
Hm, this should probably be rejected.

VCR is in play... and for some reason it doesn't work when I run the tests with `py.test` myself (which is why I made this fix..) but when I run the tests with `tox` the cassettes *do* seem to be used, and this patch breaks that.

jeremycline:
Interesting... Off the top of my head I don't know why this would be the case. Still, I'd like to support running the tests directly with py.test as well as with tox so I'll investigate this.

PR #206, opened 2017-07-12
Pagure ownership queries.

ralphbean:
This introduces a new method to query for package ownership: from
pagure-over-dist-git instead of from pkgdb.

https://fedoraproject.org/wiki/Changes/ArbitraryBranching

### DIFF
--- a/fedmsg.d/fmn.py
+++ b/fedmsg.d/fmn.py
@@ -24,6 +24,8 @@ config = {
     "datanommer.sqlalchemy.url": "postgresql+psycopg2://datanommer:bunbunbun@localhost:5432/datanommer",
 
     # Some configuration for the rule processors
+    "fmn.rules.utils.use_pagure_for_ownership": False,
+    "fmn.rules.utils.pagure_api_url": "https://src.stg.fedoraproject.org/pagure/api/",
     "fmn.rules.utils.use_pkgdb2": False,
     "fmn.rules.utils.pkgdb2_api_url": "http://209.132.184.188/api/",
     "fmn.rules.cache": {
--- a/fmn/rules/utils.py
+++ b/fmn/rules/utils.py
@@ -4,11 +4,7 @@ from collections import defaultdict
 import logging
 import time
 
-try:
-    from urllib.parse import urlencode  # python3
-except ImportError:
-    from urllib import urlencode  # python2
-
+from six.moves.urllib.parse import urlencode
 import requests
 import requests.exceptions
 
@@ -72,30 +68,38 @@ def get_fas(config):
     return _FAS
 
 
-def _paginate_pagure_data(url, params, key):
+def _paginate_pagure_data(url, params):
+    """
+    Paginate over a Pagure URL.
+
+    Args:
+        url (str): The URL to paginate over, without any parameters.
+        params (dict): A dictionary of URL parameters to use when querying Pagure.
+
+    Yields:
+        dict: The deserialized JSON response for a given page.
+    """
+
     # Set up the first page query
     params['page'] = 1
-    # Also, set the short param to make the queries lighter
-    params['short'] = 'true'
     next_page_url = url + "?" + urlencode(params)
 
-    while next_page_url is not None:
+    while next_page_url:
         try:
             response = requests.get(next_page_url, timeout=25)
-        except requests.exceptions.ConnectTimeout as e:
-            log.warn('URL %s timed out %r', response.url, e)
-            raise StopIteration
+            if response.status_code == 200:
+                data = response.json()
+                # When we run out of pages, this will be None
+                next_page_url = data['pagination']['next']
+                yield data
+            else:
+                log.error('Querying Pagure at %s returned code %s',
+                          response.url, response.status_code)
+                next_page_url = None
 
-        if not response.status_code == 200:
-            log.warn('URL %s returned code %s', response.url, response.status_code)
-            raise StopIteration
-
-        data = response.json()
-        for entry in data[key]:
-            yield entry
-
-        # When we run out of pages, this will be None
-        next_page_url = data['pagination']['next']
+        except requests.exceptions.Timeout as e:
+            log.warn('URL %s timed out %r', next_page_url, e)
+            next_page_url = None
 
 
 def get_packagers_of_package(config, package):
@@ -141,11 +145,10 @@ def _get_pagure_packagers_for(config, package):
     Returns:
         set:  A `set` of packager usernames.
     """
-    log.debug("Requesting pagure packagers of package %r" % package)
     base = config.get('fmn.rules.utils.pagure_api_url',
                       'https://src.fedoraproject.org/pagure/api')
 
-    # XXX. give a default namespace if one is not provided
+    # Give a default namespace if one is not provided. See
     # https://github.com/fedora-infra/fmn/issues/208
     if '/' not in package:
         package = 'rpms/' + package
@@ -154,8 +157,8 @@ def _get_pagure_packagers_for(config, package):
     log.info("Querying Pagure at %s for packager information", url)
     try:
         response = requests.get(url, timeout=25)
-    except requests.exceptions.ConnectTimeout as e:
-        log.warn('URL %s timed out %r', response.url, e)
+    except requests.exceptions.Timeout as e:
+        log.warn('URL %s timed out %r', url, e)
         return set()
 
     if response.status_code != 200:
@@ -192,8 +195,8 @@ def _get_pkgdb2_packagers_for(config, package):
     log.info("Querying PkgDB at %s for packager information", url)
     try:
         req = requests.get(url, timeout=15)
-    except requests.exceptions.ConnectTimeout as e:
-        log.warn('URL %s timed out %r', req.url, e)
+    except requests.exceptions.Timeout as e:
+        log.warn('URL %s timed out %r', url, e)
         return set()
 
     if not req.status_code == 200:
@@ -311,8 +314,8 @@ def _get_pkgdb2_packages_for(config, username, flags):
 
     try:
         req = requests.get(url, timeout=15)
-    except requests.exceptions.ConnectTimeout as e:
-        log.warn('URL %s timed out %r', req.url, e)
+    except requests.exceptions.Timeout as e:
+        log.warn('URL %s timed out %r', url, e)
         return set()
 
     if not req.status_code == 200:
@@ -339,42 +342,42 @@ def _get_pagure_packages_for(config, username, flags):
         username (str): The FAS username to fetch the packages for.
         flags (list): The type of relationship the user should have to the
             package (e.g. "watch", "point of contact", etc.).
+
     Returns:
-        dict:  A `dict` mapping namespaces to sets of package names.
+        dict:  A dictionary mapping namespaces to sets of package names.
+
+    Raises:
+        ValueError: If invalid flags are provided, or no flags are provided.
     """
     log.debug("Requesting pagure packages for user %r" % username)
 
+    # In the future we want to also support the 'watch' state.
     # See https://github.com/fedora-infra/fmn/issues/209
     # which depends on https://pagure.io/pagure/issue/2421
-    valid_flags = ['point of contact', 'co-maintained']#, 'watch']
+    valid_flags = ['point of contact', 'co-maintained']
 
     bogus = set(flags) - set(valid_flags)
     if bogus:
-        log.error("%r are not valid owner flags for %r." % (bogus, username))
-        # ... but, proceed.
-
-    valid = set(flags) & set(valid_flags)
-    if not valid:
-        log.error("No valid owner flags by which to query.")
-        return dict()
+        raise ValueError(
+            "{flags} are not valid owner flags for {user}.".format(flags=bogus, user=username))
+    if len(flags) == 0:
+        raise ValueError("No valid owner flags by which to query.")
 
     base = config.get('fmn.rules.utils.pagure_api_url',
                       'https://src.fedoraproject.org/pagure/api')
     url = base + '/0/projects'
 
     packages = defaultdict(set)
-    for flag in valid:
+    for flag in flags:
         if flag == 'point of contact':
-            params = dict(owner=username)
-        elif flag == 'co-maintained':
-            params = dict(username=username)
-        else:
-            # Belt and suspenders.  We should never get here.
-            raise NotImplementedError("%r is not a valid flag" % flag)
+            params = dict(owner=username, short=True)
+        else:  # 'co-maintained'
+            params = dict(username=username, short=True)
 
-        repos = _paginate_pagure_data(url, params, key='projects')
-        for repo in repos:
-            packages[repo['namespace']].add(repo['name'])
+        pages = _paginate_pagure_data(url, params)
+        for page in pages:
+            for repo in page['projects']:
+                packages[repo['namespace']].add(repo['name'])
 
     return dict(packages)
 
--- a/fmn/rules/utils.py
+++ b/fmn/rules/utils.py
@@ -4,6 +4,11 @@ from collections import defaultdict
 import logging
 import time
 
+try:
+    from urllib.parse import urlencode  # python3
+except ImportError:
+    from urllib import urlencode  # python2
+
 import requests
 import requests.exceptions
 
@@ -67,6 +72,32 @@ def get_fas(config):
     return _FAS
 
 
+def _paginate_pagure_data(url, params, key):
+    # Set up the first page query
+    params['page'] = 1
+    # Also, set the short param to make the queries lighter
+    params['short'] = 'true'
+    next_page_url = url + "?" + urlencode(params)
+
+    while next_page_url is not None:
+        try:
+            response = requests.get(next_page_url, timeout=25)
+        except requests.exceptions.ConnectTimeout as e:
+            log.warn('URL %s timed out %r', response.url, e)
+            raise StopIteration
+
+        if not response.status_code == 200:
+            log.warn('URL %s returned code %s', response.url, response.status_code)
+            raise StopIteration
+
+        data = response.json()
+        for entry in data[key]:
+            yield entry
+
+        # When we run out of pages, this will be None
+        next_page_url = data['pagination']['next']
+
+
 def get_packagers_of_package(config, package):
     """ Retrieve the list of users who have commit on a package.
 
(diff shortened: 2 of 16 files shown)


fmn/rules/utils.py
```diff
@@ -79,10 +79,49 @@ def get_packagers_of_package(config, package):
...
+    if not response.status_code == 200:
+        log.warn('URL %s returned code %s', response.url, response.status_code)
+        return set()
+
+    data = response.json()
+
+    packagers = set(sum(data['access_users'].values(), []))
+    groups = set(sum(data['access_groups'].values(), []))
+
+    # XXX - Should we get group membership from pagure instead of FAS?  Ask pingou.
```
ralphbean: @pypingou, any thoughts here?

pypingou: So the idea is that group membership will be synced from FAS to pagure, so FAS should always be up to date while pagure may not (you'd need to log-out and relog-in for your membership to be refreshed).
@puiterwijk said he would be looking into confirming this works.

So on the one side FAS is correct and always up to date, on the other side this means people may receive notifications about, say PR that they wouldn't be able to act on (not without that logout/login dance).
Kinda still inclined for FAS though.

jeremycline: How long is a user allowed to stay logged in for? I do see the appeal of not depending on FAS's API, but I agree it wouldn't be great to have a huge lag before you start getting notifications.

pypingou: They can stayed logged in for quite a while, they'll need to relogin for every access to the settings page of a project made 20 minutes or more from their login time, otherwise, their session lasts as long as flask's default which is iirc 31 days.

fmn/rules/utils.py
```diff
@@ -79,10 +79,49 @@ def get_packagers_of_package(config, package):
         _cache.configure(**config['fmn.rules.cache'].copy())
 
     key = cache_key_generator(get_packagers_of_package, package)
-    creator = lambda: _get_pkgdb2_packagers_for(config, package)
+    creator = lambda: _get_packagers_for(config, package)
     return _cache.get_or_create(key, creator)
 
 
+def _get_packagers_for(config, package):
```
jeremycline, line 121: Could you please add a docblock here?

fmn/rules/utils.py
```diff
@@ -79,10 +79,49 @@ def get_packagers_of_package(config, package):
...
+    if config['fmn.rules.utils.use_pagure_for_ownership']:
+        return _get_pagure_packagers_for(config, package)
+    else:
+        return _get_pkgdb2_packagers_for(config, package)
+
+
+def _get_pagure_packagers_for(config, package):
```
jeremycline, line 139: Could you please add a docblock here?

fmn/rules/utils.py
```diff
@@ -79,10 +79,49 @@ def get_packagers_of_package(config, package):
...
+    log.debug("Requesting pagure packagers of package %r" % package)
+    default = 'https://src.fedoraproject.org/pagure/api'
+    base = config.get('fmn.rules.utils.pagure_api_url', default)
+
+    # XXX. give a default namespace if one is not provided
+    # Later, we need to come back in here and make all of this namespace-aware.
```
jeremycline: Can you file this as an issue that points to this code and remove the to-do from the code?

fmn/rules/utils.py
```diff
@@ -79,10 +79,49 @@ def get_packagers_of_package(config, package):
...
+    if '/' not in package:
+        package = 'rpms/' + package
+
+    url = '{0}/0/{1}'.format(base, package)
+    log.info("hitting url: %r" % url)
```
jeremycline: It would be nice if this was a little more explicit so it's clear why the URL is being queried. Something like ``log.info("Querying Pagure at %s for packager information", url)``

fmn/rules/utils.py
```diff
@@ -79,10 +79,49 @@ def get_packagers_of_package(config, package):
...
+    response = requests.get(url)
```
jeremycline: I filed https://github.com/fedora-infra/fmn/issues/207 about using a session, but could you add a timeout to this call? ``timeout=15`` should be fine. It should be fine to handle the exception here and just do what you're doing when the return code isn't 200.

fmn/rules/utils.py
```diff
@@ -79,10 +79,49 @@ def get_packagers_of_package(config, package):
...
+
+    if not response.status_code == 200:
```
jeremycline: ``if response.status_code != 200``, maybe?

fmn/rules/utils.py
jeremycline: I guess a month is a bit long to wait before getting notifications. I guess I also vote sticking with FAS for the moment.

fmn/rules/utils.py
```diff
@@ -162,6 +201,13 @@ def invalidate_cache_for(config, fn, arg):
     return _cache.delete(key)
 
 
+def _get_packages_for(config, username, flags):
```
jeremycline, line 272: Docblock here please 😃

fmn/rules/utils.py
```diff
@@ -198,6 +244,58 @@ def _get_pkgdb2_packages_for(config, username, flags):
...
+
+    Args:
+        config (dict): The application configuration.
+        username (str): The FAS username to fetch the packages for.
+        flags (list): The type of relationship the user should have to the
+            package (e.g. "watch", "point of contact", etc.).
+    """
+    log.debug("Requesting pagure packages for user %r" % username)
+
+    # XXX - We don't yet have a way to query for all projects that a user
```
jeremycline: Could you file this as an issue that points to this and remove the TODO from the code?

fmn/rules/utils.py
```diff
@@ -198,6 +244,58 @@ def _get_pkgdb2_packages_for(config, username, flags):
...
 
+def _get_pagure_packages_for(config, username, flags):
+    """
+    Get the packages a user is associated with from pagure.
+
+    Args:
+        config (dict): The application configuration.
+        username (str): The FAS username to fetch the packages for.
+        flags (list): The type of relationship the user should have to the
+            package (e.g. "watch", "point of contact", etc.).
```
jeremycline, line 344: Could you add the return value here?
```
Returns:
    <type>: Description of return
```

fmn/rules/utils.py
```diff
@@ -198,6 +244,58 @@ def _get_pkgdb2_packages_for(config, username, flags):
...
+    if bogus:
+        log.error("%r are not valid owner flags for %r." % (bogus, username))
+        # ... but, proceed.
+
+    valid = set(flags) & set(valid_flags)
+    if not valid:
+        log.error("No valid owner flags by which to query.")
+        return dict()
+
+    default = 'https://src.fedoraproject.org/pagure/api'
```
jeremycline: Since this is never referenced after its single use directly below, you could probably just use it directly instead of assigning it to a variable

fmn/rules/utils.py
```diff
@@ -198,6 +244,58 @@ def _get_pkgdb2_packages_for(config, username, flags):
...
+    for flag in valid:
+        if flag == 'point of contact':
+            params = dict(owner=username)
+        elif flag == 'co-maintained':
+            params = dict(username=username)
+        else:
+            # Belt and suspenders.  We should never get here.
+            raise NotImplementedError("%r is not a valid flag" % flag)
+
+        response = requests.get(url, params=params)
```
jeremycline: Same deal with the timeout here as above

jeremycline (changes requested):
Thanks for this, it looks really solid. I just have a few change requests.

ralphbean:
Thanks @jeremycline, 75cf3c7 should address all your comments.

fedmsg.d/fmn.py
```diff
@@ -24,6 +24,8 @@
     "datanommer.sqlalchemy.url": "postgresql+psycopg2://datanommer:bunbunbun@localhost:5432/datanommer",
 
     # Some configuration for the rule processors
+    "fmn.rules.utils.use_pagure_for_ownership": False,
+    "fmn.rules.utils.pagure_api_url": "https://src.stg.fedoraproject.org/pagure/api/",
```
mprahl, line 28: Did you want to put the prod URL even though it isn't deployed yet?

fmn/rules/utils.py
```diff
@@ -79,18 +79,91 @@ def get_packagers_of_package(config, package):
...
+    url = '{0}/0/{1}'.format(base, package)
+    log.info("Querying Pagure at %s for packager information", url)
+    try:
+        response = requests.get(url, timeout=15)
+    except requests.exceptions.ConnectTimeout as e:
+        log.warn('URL %s timed out %r', response.url, e)
+        return set()
+
+    if response.status_code != 200:
+        log.warn('URL %s returned code %s', response.url, response.status_code)
```
mprahl: Optional, but it might be nice to add `response.text` in the log statement if it's a 500 error.

fmn/rules/utils.py
```diff
@@ -79,18 +79,91 @@ def get_packagers_of_package(config, package):
...
+        return set()
+
+    data = response.json()
+
+    packagers = set(sum(data['access_users'].values(), []))
```
mprahl, line 170: Are you sure you want to include people who just have ticket access in this? I assume you only want admins and committers.

fmn/rules/utils.py
```diff
@@ -79,18 +79,91 @@ def get_packagers_of_package(config, package):
...
+    groups = set(sum(data['access_groups'].values(), []))
```
mprahl, line 171: Same as above.

fmn/rules/utils.py
```diff
@@ -162,6 +235,27 @@ def invalidate_cache_for(config, fn, arg):
     return _cache.delete(key)
 
 
+def _get_packages_for(config, username, flags):
```
mprahl, line 272: This seems very similar to `fmn.rules.utils. _get_packagers_for`. Is there a way to consolidate these?

Edit:
NVM, I misread the function names :smile:

ralphbean: They are.. opposites of each other?  One gets packagers for a package - the other gets packages for a packager.

fedmsg.d/fmn.py
ralphbean: The example pkgdb url given in the same file is *also* a non-prod url, so I went with that.

fmn/rules/utils.py
ralphbean: (Done).

fmn/rules/utils.py
```diff
@@ -79,18 +79,91 @@ def get_packagers_of_package(config, package):
...
+    Returns:
+        set:  A `set` of packager usernames.
+    """
+    if config['fmn.rules.utils.use_pagure_for_ownership']:
+        return _get_pagure_packagers_for(config, package)
+    else:
+        return _get_pkgdb2_packagers_for(config, package)
+
+
+def _get_pagure_packagers_for(config, package):
```
ralphbean, line 139: (Done).

fmn/rules/utils.py
ralphbean: Nope - them too.  This is for routing notifications to packagers about packages "they are interested in"... and the old pkgdb logic here included everyone with any kind of association with the package.

fmn/rules/utils.py
ralphbean: (Done).

fmn/rules/utils.py
ralphbean: (Done).

fmn/rules/utils.py
```diff
@@ -198,6 +299,65 @@ def _get_pkgdb2_packages_for(config, username, flags):
...
+    url = base + '/0/projects'
+
+    packages = defaultdict(set)
+    for flag in valid:
+        if flag == 'point of contact':
+            params = dict(owner=username)
+        elif flag == 'co-maintained':
+            params = dict(username=username)
+        else:
+            # Belt and suspenders.  We should never get here.
```
mprahl: lol

fmn/rules/utils.py
```diff
@@ -198,6 +299,65 @@ def _get_pkgdb2_packages_for(config, username, flags):
...
+        # ... but, proceed.
+
+    valid = set(flags) & set(valid_flags)
+    if not valid:
+        log.error("No valid owner flags by which to query.")
+        return dict()
+
+    base = config.get('fmn.rules.utils.pagure_api_url',
+                      'https://src.fedoraproject.org/pagure/api')
+    url = base + '/0/projects'
```
mprahl: You may want to consider using pagination (you have to specify a page value for pagination to be turned on):
https://src.stg.fedoraproject.org/pagure/api/0/projects?page=1&username=ralph

fmn/rules/utils.py
```diff
@@ -198,6 +299,65 @@ def _get_pkgdb2_packages_for(config, username, flags):
...
+        if flag == 'point of contact':
+            params = dict(owner=username)
+        elif flag == 'co-maintained':
+            params = dict(username=username)
+        else:
+            # Belt and suspenders.  We should never get here.
+            raise NotImplementedError("%r is not a valid flag" % flag)
+
+        try:
+            response = requests.get(url, params=params, timeout=15)
```
mprahl: If you end up not using pagination, you should increase this timeout.

fmn/rules/utils.py
```diff
@@ -198,6 +299,65 @@ def _get_pkgdb2_packages_for(config, username, flags):
...
+        except requests.exceptions.ConnectTimeout as e:
+            log.warn('URL %s timed out %r', response.url, e)
+            continue
+
+        if not response.status_code == 200:
+            log.warn('URL %s returned code %s', response.url, response.status_code)
```
mprahl: It may be nice to log `response.text` here.

fmn/rules/utils.py
mprahl: Actually, since it looks like you just want the project names, you could add the `short=true` parameter which only gives a small subset of info per project which is much faster.

fmn/rules/utils.py
```diff
@@ -198,6 +299,65 @@ def _get_pkgdb2_packages_for(config, username, flags):
...
+        flags (list): The type of relationship the user should have to the
+            package (e.g. "watch", "point of contact", etc.).
+    Returns:
+        dict:  A `dict` mapping namespaces to sets of package names.
+    """
+    log.debug("Requesting pagure packages for user %r" % username)
+
+    # See https://github.com/fedora-infra/fmn/issues/209
+    # which depends on https://pagure.io/pagure/issue/2421
+    valid_flags = ['point of contact', 'co-maintained']#, 'watch']
```
mprahl: Can you remove this comment at the end of this line?

fmn/rules/utils.py
mprahl: Yeah, I edited my comment as you replied. Just reading issues on my end.

mprahl:
@ralphbean just some minor comments, particularly about being more efficient using the Pagure API. It's up to you if you want to address these comments or not.

+1 from me either way.

fmn/rules/utils.py
ralphbean: Aw... really?  It will make the edit to re-enable it so simple.  :p

mprahl: Sorry, it confused me when first reading. I didn't see that it was commented out. 😛

fmn/rules/utils.py
jeremycline: I'd say if the API supports pagination it'd be good to make use of it. I don't know how many users have thousands of packages in their point of contact or co-maintained list, but I'd prefer to do it the right way before hitting a problem in production.

Are there already Python bindings out there that handle the pagination? It seems like something everyone would want to (re)use.

mprahl: @jeremycline I agree. There aren't any Python bindings yet since the PR was just merged 6 days ago:
https://pagure.io/pagure/pull-request/2417

I don't have any plans on writing any but I'm sure many would appreciate it.

ralphbean: FYI - I go on PTO end of day Friday.  I granted you both access to my fork in case you want to push additional patches there to update this PR:  https://github.com/ralphbean/fmn

I'll try to fix before I leave!

jeremycline: @ralphbean Just so you know, if you tick the "Allow edits from maintainers" box when making a PR (the default), anyone with write access to the repository can push to the branch in your fork :)

ralphbean: > if you tick the "Allow edits from maintainers" box when making a PR (the default)

I did *not* know this.  Cool.

Well - it is ticked.  :)

fmn/rules/utils.py
ralphbean: Pagination added in b6ac6a0768ee0b59ca6814bede58aa4ed6189cee

fmn/rules/utils.py
ralphbean: OK, pagination added in b6ac6a0768ee0b59ca6814bede58aa4ed6189cee.

ralphbean:
> being more efficient using the Pagure API

Pagure pagination added in b6ac6a0768ee0b59ca6814bede58aa4ed6189cee.

fmn/rules/utils.py
```diff
@@ -341,18 +370,8 @@ def _get_pagure_packages_for(config, username, flags):
...
-            log.warn('URL %s timed out %r', response.url, e)
-            continue
-
-        if not response.status_code == 200:
-            log.warn('URL %s returned code %s', response.url, response.status_code)
-            continue
-
-        data = response.json()
-        for repo in data['projects']:
+        repos = _paginate_pagure_data(url, params, key='projects')
```
mprahl: Can you add `short=true` to the parameters since you only need the namespace and name of project? It will speed up the queries significantly.

fmn/rules/utils.py
```diff
@@ -67,6 +72,30 @@ def get_fas(config):
...
 
 
+def _paginate_pagure_data(url, params, key):
+    # Set up the first page query
+    params['page'] = 1
+    next_page_url = url + "?" + urlencode(params)
+
+    while next_page_url is not None:
+        try:
+            response = requests.get(next_page_url, timeout=15)
```
mprahl: With Pagure being inconsistent with speeds, you may want to bump up the timeout.

mprahl:
@ralphbean small comment. Overall, I'm +1 on this change but it seems that the tests are failing. I don't think they are your fault though.

fmn/rules/utils.py
ralphbean: Done in b8ba186

fmn/rules/utils.py
ralphbean: Done in 148ea66.

ralphbean:
> ...it seems that the tests are failing. I don't think they are your fault though.

Agreed.  Those fail even on the develop branch.

mprahl:
@jeremycline I don't have rights to merge this. Are you good with the changes?

jeremycline:
@mprahl I'm going to investigate and fix the test failures before I merge this. The PR looks fine, though.

fmn/rules/utils.py
```diff
@@ -4,6 +4,11 @@
 import logging
 import time
 
+try:
```
jeremycline: You should be able to just use ``six`` for this.

fmn/rules/utils.py
```diff
@@ -67,6 +72,32 @@ def get_fas(config):
     return _FAS
 
 
+def _paginate_pagure_data(url, params, key):
```
jeremycline: This needs a documentation block, including the exceptions it raises.

jeremycline (changes requested):

fmn/rules/utils.py
jeremycline: Okay, I've documented this API, but I think we should change it.

For one thing, I don't like that errors are quietly swallowed. Things using this API may want to make decisions based on these errors like whether or not to cache the response. As a side note, the behavior of generators raising ``StopIteration`` [is about to change in Python 3.7](https://docs.python.org/3/library/exceptions.html#StopIteration). I only skimmed the associated [PEP](https://www.python.org/dev/peps/pep-0479/), but it seems like what is being done here is not correct (not 100% certain about that, though).

I also find the interface of passing a URL, the URL parameters, and a key I want from a JSON response odd. Why not just return the entire JSON object? Anyone who wants more than one key from the response is not going to enjoy using this.

I'd prefer to use pagination, but I'd also prefer us to not "roll our own" here when we could do it once in a Python binding library. @pypingou, is there an "official" Python bindings library?

pypingou: There is an effort that was started around http://pagure.io/libpagure it's likely the closest thing to something official, but I do not know how updated it is.

jeremycline: Hmm. Given how many applications are going to start using the API from Python (Bodhi, this, the-new-hotness, etc.), it might be worth-while to concentrate everyone's efforts on a Python interface. Maybe something in the pagure repository itself, as [bodhi does with its client](https://github.com/fedora-infra/bodhi/tree/2.8/bodhi/client)?

mprahl: @jeremycline I think it'd be helpful to have a shared library, but that's really a separate discussion. Can we modify FMN to use the library once it has been written, so it doesn't block this PR?

jeremycline: Very well. I filed #213 (and an associated Pagure issue) to track that. I still don't think the current API here makes sense, though.

I'm guessing there's a time crunch here. When does this need to be in production?

bowlofeggs: Bodhi does have some code from @mprahl to make API calls to Pagure - I would love to use an official Python binding library if such a thing were to be created. It's not a huge deal though, as Bodhi only really makes "light" usage of Pagure so far.

https://github.com/fedora-infra/bodhi/blob/2.8.1/bodhi/server/util.py#L775-L781

pypingou: Pagure is due to be deployed when the mass-rebuild finished committing to dist-git, so either the end of the week or the beginning of next week.

pkgdb will stick around a little longer, but hopefully not more than a week or two, time for things to be ported/deployed.

mprahl: @jeremycline is this the only feedback left to be addressed?

> I also find the interface of passing a URL, the URL parameters, and a key I want from a JSON response odd. Why not just return the entire JSON object? Anyone who wants more than one key from the response is not going to enjoy using this.

jeremycline: No, the test coverage for the diff needs to be 100%. I've refactored this PR locally to make the API more sensible and written most of the tests, but the staging Pagure instance seems to have issues with packages with groups as committers or something: e.g. https://src.stg.fedoraproject.org/pagure/api/0/rpms/python-requests. I can't finish the tests without this working.

ralphbean: @pypingou, do you know what @jeremycline is talking about here?

> the staging Pagure instance seems to have issues with packages with groups as committers or something

Jeremy, do you have more details about what's failing here?

jeremycline: pingou fixed it late on Friday, but I don't know what the issue was.

pypingou: The root cause was that pagure doesn't allow a group as main maintainer of a project so the sync script for failing, it was saying so in the logs but the logs are verbose so I missed it until @jeremycline pointed it out.
So I went with: if the PoC is a group, use as main admin the first person in the project who is not a group and we will need to use the override repo to put back in place the group as PoC in BZ.

ralphbean (commented):
LGTM.

jeremycline (approved):

PR #211, opened 2017-07-21
Set up datanommer in the dev environment

jeremycline:

### DIFF
--- a/ansible/roles/fmn-dev/tasks/db.yml
+++ b/ansible/roles/fmn-dev/tasks/db.yml
@@ -1,0 +1,72 @@
+---
+- name: Install database packages
+  dnf:
+      name: "{{ item }}"
+      state: present
+  with_items:
+      - postgresql-server
+      - postgresql-devel
+      - postgresql
+      - python-psycopg2
+
+- name: Install datanommer
+  dnf:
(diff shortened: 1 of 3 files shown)


ansible/roles/fmn-dev/tasks/db.yml
```diff
@@ -0,0 +1,72 @@
...
+    replace: "host    all             all             127.0.0.1/32            trust"
+
+- replace:
+    dest: /var/lib/pgsql/data/pg_hba.conf
+    regexp: "host    all             all             ::1/128                 ident"
+    replace: "host    all             all             ::1/128                 trust"
+
+
+- name: Start and enable postgresql
+  service: name=postgresql state=restarted enabled=yes
```
ralphbean, line 38: Do you need/want to also start and enable datanommer (fedmsg-hub)?

jeremycline: At the moment none of the services get started in the VM. There's an alias in the .bashrc to start/stop/restart everything (including fedmsg-hub). I think I made it that way because some config needs to be tweaked (FAS account stuff to query the API, for example) before you can start things.

ralphbean (approved):
:hamburger:

PR #212, opened 2017-07-24
Pin urllib3 and requests

jeremycline:
The two test failures were due to fedmsg_meta quietly failing to load
processors. This failed because at some point at some point during the
processor loading phase it was importing requests which was upset about
the urllib3 version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,9 +19,9 @@ python-openid
 python-openid-cla
 python-openid-teams
 redis
-requests
+urllib3>=1.21.1,<1.22
+requests>=2.18.1
 six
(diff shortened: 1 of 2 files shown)


puiterwijk:
Hmm, hardcoding to a particular version of a dependency is a bad idea.
Do you maybe want a >= or something?

puiterwijk:
Since this is just for CI, +1.
Just make sure to pull it out before any release.

jeremycline:
Thanks. Just for those who might stumble upon this later, the pinning is so that CI will pass until requests releases a new version that increases its urllib3 compatibility range. This will be backed out in a couple days.

PR #218, opened 2017-08-07
Make the topic the FMN consumer subscribes to configurable

### LINKED ISSUE #216: Allow the topic fmn listens to to be configurable
FMN currently listens to everything (``b''``) on all the ZeroMQ sockets it connects to. It should be configurable so we filter for production (``b'org.fedoraproject.prod'`` for example).

This should accept either a unicode string or list of unicode strings.

jeremycline:
Allow the user to configure what topic(s) the FMN consumer subscribes
to. The default remains "everything".

Fixes #216


pypingou:
Cool 👍 for me thanks!

PR #221, opened 2017-08-10
Start adding CI related rules

pypingou:
Signed-off-by: Pierre-Yves Chibon <pingou@pingoured.fr>

### DIFF
--- a/fmn/rules/ci.py
+++ b/fmn/rules/ci.py
@@ -1,0 +1,235 @@
+from fmn.lib.hinting import hint, prefixed as _
+
+# Basic building steps
+
+@hint(categories=['ci'])
+def ci_catchall(config, message):
+    """ All CI events
+
+    Adding this rule will indiscriminately match notifications of all types
+    from the `Fedora Atomic CI pipeline
+    <https://fedoraproject.org/wiki/FedoraAtomicCI/pipeline>`_, i.e.
+    new build queued, running, complete, compose running, complete, etc..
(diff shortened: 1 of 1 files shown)


fmn/rules/ci.py
```diff
@@ -0,0 +1,234 @@
+from fmn.lib.hinting import hint, prefixed as _
+
+# Basic building steps
+
+@hint(categories=['ci'])
+def ci_catchall(config, message):
+    """ All CI events
+
+    Adding this rule will indiscriminately match notifications of all types
+    from `ci <https://admin.fedoraproject.org/pkgdb>`_, i.e. new build queued,
```
ralphbean: This URL looks... wrong.

ralphbean: Same url appears elsewhere in the module.  Maybe use https://fedoraproject.org/wiki/FedoraAtomicCI/pipeline ?

You may always want to replace the vague "ci" with the more specific "the Fedora Atomic CI pipeline".

ralphbean (changes requested):

fmn/rules/ci.py
pypingou: Yeah, I wasn't sure what would be a good URL to use, the wiki page? the actual jenkins instance?

Maybe the wiki page is better since the jenkins instance will be linked to in the messages.

jeremycline (approved):
Thanks!

PR #224, opened 2017-08-16
Fix showing the CI rules in the UI

pypingou:


jeremycline:
To fix the tests, there are two options. Adjust the test to make sure the mocked message has a 'msg' key or adjust the rules so it returns False if there's no 'msg' key. The test as it currently stands seems to indicate that there could be messages without the 'msg' body, but I have no idea about the validity of this because there's no message schemas.

I lean making your rules return False if there's no 'msg' key simply because everything else is so fragile and will likely explode if that's ever not true for some reason.

pypingou:
Note that I took the example of on the of the taskotron rules to check this (wasn't sure if the message that was passed had a ``msg`` or if I was already inside it.

pypingou:
👍 for me w/ your changes :)

pypingou:
Hm, I thought invertible was used to negate a rule (ie go from: "all CI messages about a commit being ignored" to: "all CI messages except the ones about a commit being ignored") if that is the case then I think we still may want to allow to negate these rules

jeremycline:
Well, there's no documentation so I can only guess that that's the case without observing how it's used everywhere. You're welcome to attempt to figure out what the API actually is so that the test passes, but I have a hard time imagining there's a large group of people out there who want to get _all_ messages except for the CI-related messages.

pypingou:
Explanation from @ralphbean:
~~~~
threebean | the invertible flag tells @hint how closely the hint matches the real rule.
threebean | so if the rule returns true if category == 'foo' and the hint says category='foo'...
threebean | ...then they matche exactly, and the hint is invertible.  if you negate the hint, it gives the same results as negating the rule.
threebean | but some rules are like category == 'foo' and user == 'pingou', but the hint is only for category='foo'
threebean | in this case the hint is not invertible.
threebean | (the hint is a hint to a datanommer query for constructing the "examples" query in fmn.web)
~~~~

pypingou:
👍 for me

PR #228, opened 2017-09-08
Enable the IRC client to authenticate with NickServ

### LINKED ISSUE #217: FMN doesn't deliver messages through matrix-irc bridge
After I switched for IRC to use Matrix, I stopped getting any fedmsg notifications which is weird..

@nirik suggested to open ticket here one day.u

jeremycline:
This also adds the ability to turn TLS on for the IRC connection.

As is always there case with FMN, there aren't any tests for this. Pretty much all of the backend needs to be re-written anyway. This particular issue is more urgent since Freenode enabled +R for everyone by default.

fixes #217 


pypingou:
Looks fine to me :)

PR #231, opened 2017-09-15
Convert the workers to Celery

jeremycline:
This changes the worker processes to use Celery. All queue names and message formats remain unchanged (except for a few AMQP header values). The fedmsg consumer has also been converted to use Kombu rather than Pika for AMQP messages as it's a dependency of Celery and is a much higher-level API.


ansible/roles/fmn-dev/files/fmn-worker.service
```diff
@@ -1,11 +0,0 @@
-[Unit]
```
bowlofeggs, line 1: Why not keep it called ```fmn-worker.service```? Celery is just an implementation detail.

fmn/config.py
```diff
@@ -0,0 +1,6 @@
+"""This module is responsible for loading the application configuration."""
+import fedmsg
+
+
+#: The application configuration dictionary.
+app_conf = fedmsg.config.load_config()
```
bowlofeggs, line 6: I recommend against having the import of this module cause config loading to happen. It makes testing tricky. IIRC, Celery has events you can use to load the config during process start up.

systemd/fmn-worker@.service
```diff
@@ -1,14 +0,0 @@
-[Unit]
```
bowlofeggs: I see you are taking this fmn-worker service away, but do you want to provide an example Celery service file in its place? If so, I suggest keeping the existing name.

bowlofeggs (approved):

ansible/roles/fmn-dev/files/fmn-worker.service
jeremycline: It turns out the celery workers do not like the missing fields from the pre-celery unprocessed messages (no task ID, for example). I guess it doesn't really matter in the development environment, but some sort of careful migration strategy is necessary that involves shutting down old services and starting new services as queues drain. I'm going to write up docs as soon as I figure out a good approach and test it thoroughly.

fmn/config.py
jeremycline: Totally agree. I started writing a lazy-loaded config, but that would have spiraled out of control. Before, every module was loading the config this way separately. I figured this would be a nicer middle ground where I don't add to that, but I also don't have to audit to code for the dozens of places configuration is being loaded.

It's going to happen, I just am not sure if it'll happen after I re-write the backends or before.

systemd/fmn-worker@.service
jeremycline: So I've added a replacement, although I do expect it to change as I fiddle with celery more.

PR #232, opened 2017-09-28
Switch to Celery broadcasts for cache invalidation

jeremycline:
Workers need to refresh their caches when we get messages about user
preference changes. This sets up a broadcast for workers.

There are a few oddities with this patch. Firstly, because of the way
the cache is initialized (inside the Task instance) we have to either dig
around in there to update the cache or dispatch the task itself and do a
special check to see if it's a cache refresh task. In this patch I've
opted for the second approach, although I'm not thrilled about that. In
the future we should have a better caching layer that makes this less
painful. Secondly, workers are being started in --pool=solo mode (one
process per worker). This is because the broadcast gets sent to each
worker, but not to each process in the worker's process pool. Running
in solo mode maps workers to processes 1-to-1, although it is generally
less efficient to run that way.

### DIFF
--- a/fedmsg.d/fmn.py
+++ b/fedmsg.d/fmn.py
@@ -93,7 +93,12 @@ config = {
         'include': ['fmn.tasks'],
         'accept_content': ['json'],
         'task_serializer': 'json',
-        'task_default_queue': 'workers',
+        'task_routes': {
+            'fmn.tasks.find_recipients': {
+                'queue': 'fmn.tasks.unprocessed_messages',
+                'routing_key': 'fmn.tasks.unprocessed_messages',
+            },
+        }
     },
 
     # Generic stuff
--- a/fmn/consumer/consumer.py
+++ b/fmn/consumer/consumer.py
@@ -12,70 +12,24 @@ queries to `FAS`_, `pkgdb`_, or the database.
 .. _pkgdb: https://admin.fedoraproject.org/pkgdb/
 """
 
-import datetime
 import logging
-import uuid
 
-from kombu import Connection, Exchange
-from kombu.pools import connections
 import fedmsg.consumers
 import kombu
 
 import fmn.lib
 import fmn.rules.utils
-from fmn import config
+from fmn.celery import RELOAD_CACHE_EXCHANGE_NAME
 from fmn.consumer.util import (
     new_packager,
     new_badges_user,
     get_fas_email,
 )
-from fmn.tasks import find_recipients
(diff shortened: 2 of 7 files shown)


fedmsg.d/fmn.py
```diff
@@ -93,7 +93,12 @@
         'include': ['fmn.tasks'],
         'accept_content': ['json'],
         'task_serializer': 'json',
-        'task_default_queue': 'workers',
+        'task_routes': {
+            'fmn.tasks.find_recipients': {
+                'queue': 'fmn.tasks.unprocessed_messages',
+                'routing_key': 'fmn.tasks.unprocessed_messages',
```
bowlofeggs, line 99: This change looks quite similar to the change made in ```ansible/roles/fmn-dev/files/fedmsg.d/fmn.py``` - is there an opportunity to share code here?

fmn/consumer/consumer.py
```diff
@@ -212,7 +169,11 @@ def work(self, session, raw_msg):
                 )
                 session.add(user)
                 session.commit()
-                notify_prefs_change(openid)
+                _log.info('Broadcasting message to Celery workers to update cache for %s', openid)
+                find_recipients.apply_async(
+                    ({'topic': 'fmn.internal.refresh_cache', 'body': openid},),
+                    exchange='fmn.tasks.reload_cache'
```
bowlofeggs: This string gets used in several places - might make sense to make it a "constant".

fmn/tasks.py
```diff
@@ -138,11 +138,13 @@ def run(self, message):
...
         _log.debug('Determining recipients for message "%r"', message)
-
         topic, message_body = message['topic'], message['body']
-        if '.fmn.' in topic:
-            openid = message_body['msg']['openid']
-            fmn_lib.update_preferences(openid, self.user_preferences)
+
+        # We send a fake message with this topic as a broadcast to all workers in order for them
+        # to refresh their caches, so if this message is a cache refresh notification stop early.
+        if topic == 'fmn.internal.refresh_cache':
```
bowlofeggs: This might also make sense to turn into a varaible that can be imported and used.

bowlofeggs (approved):
LGTM.

fedmsg.d/fmn.py
jeremycline: Yes, but I'd prefer to put that off. My next PR will be refactoring the backend subpackage, followed by a configuration module with defaults and all that jazz. The next release is going to be fmn-2.0 and I'm going to take the opportunity to clean up the configuration.

PR #235, opened 2017-10-13
Dispatch a task on consumer startup to cache all of FAS

jeremycline:
We need everyone's email/irc/groups to do message classification. This
is fetched lazily as necessary, but we know we're going to need nearly
everyone's account information anyway, so dispatches a task right at
startup to populate the cache in the background.


pypingou (approved):
LGTM

PR #238, opened 2017-10-19
Just nack and requeue the message on failure in delivery

jeremycline:
Previously, the delivery service formatted the messages and regularly
hit messages it couldn't render due to issues with fedmsg.meta. Now,
however, all the message formatting (except for confirmations in the IRC
backend which don't use fedmsg.meta) occurs on the workers when the
message recipients are calculated. This means that the delivery service
should only fail when a) there's a bug in the code where messages can't
be delivered at all or b) there's a network partition so we're not
connected to the mail server/irc/etc.

Because of this, we can stop dequeuing and manually republishing the
message to make it go to the back of the queue. In theory, any exception
should occur regardless of the message. It would be even better to
separate delivery backend messages into separate queues so that if IRC
is unreachable, it doesn't congest email.

Depends on #237 

### DIFF
--- a/fmn/delivery/service.py
+++ b/fmn/delivery/service.py
@@ -33,7 +33,6 @@ import logging
 
 from kombu import Connection, Queue
 from kombu.mixins import ConsumerMixin
-from kombu.pools import connections
 from twisted.application import service
 from twisted.internet import threads, reactor, task, defer
 import six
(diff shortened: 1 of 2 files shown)


pypingou:
Does requeue do the right thing and push the message to the back of the queue (not the front)?
Otherwise we may quickly end up in an infinite loop.

puiterwijk (approved):

puiterwijk:
@pypingou pretty sure that this queues at the front of the queue, but if the fedmsg.meta stuff happens in the workers now, that's fine.
As Jeremy explained, if a message fails in this part, it should impact all messages of that type until it gets fixed.

jeremycline:
> @pypingou pretty sure that this queues at the front of the queue, but if the fedmsg.meta stuff happens in the workers now, that's fine.

This is correct, it'll be placed back at the front of the queue. The one (minor) downside to this approach at the moment is that there's a single backend queue at the moment so a problem with the irc network will hold up email. We'll need to start publishing the backend messages using routing keys based on the backend type to solve that, but that's easy.

PR #239, opened 2017-10-19
Add defaults to the FMN configurations

jeremycline:
This is the first step to having easy-to-understand configuration in FMN. It gathers the various configuration options (that I could find) into a single module, sets defaults, and adds the skeleton for easy validation (inspired by @bowlofeggs work with Bodhi).

It doesn't focus on writing the validation functions just yet, nor does it try to set reasonable defaults to the configuration or remove duplicate configuration.


fmn/config.py
```diff
@@ -1,6 +1,341 @@
...
+
+    def pop(self, *args, **kw):
+        if not self.loaded:
+            self.load_config()
+        return super(FmnConfig, self).pop(*args, **kw)
+
+    def copy(self, *args, **kw):
+        if not self.loaded:
+            self.load_config()
+        return super(FmnConfig, self).copy(*args, **kw)
```
bowlofeggs: It'd be nice to give the above group of methods docblocks.

fmn/config.py
```diff
@@ -1,6 +1,341 @@
...
+            self.update(settings)
+        else:
+            self.update(fedmsg.config.load_config())
+        self.loaded = True
+        self._validate()
+
+    def _load_defaults(self):
+        """Iterate over self._defaults and set all default values on self."""
+        for k, v in self._defaults.items():
+            self[k] = v['value']
```
bowlofeggs: I'd kinda wished I'd named this key "default" instead of "value". Maybe I'll change that in Bodhi one day. Just a thought.

fmn/config.py
```diff
@@ -1,6 +1,341 @@
...
+
+    def _validate(self):
+        """Run the validators found in self._defaults on all the corresponding values."""
```
bowlofeggs: Might be good to document that it raises ```ValueError```.

bowlofeggs (approved):

fmn/config.py
jeremycline: I shall learn from your regrets. I'll make it better, stronger. I have the technology.

PR #241, opened 2017-10-21
Use one queue per messaging backend

jeremycline:
Publish messages to the backend using a routing key based off the
backend (``fmn.backend.`` + ``<backend-name>``). This solves the issue
introduced in #238 where the irc network being down causes emails to
back up. The delivery service round-robins the various queues.

cc @puiterwijk  @pypingou


bowlofeggs (approved):
Looks fine to me, though the CI tests seem to fail so I recommend sorting those out.

PR #246, opened 2017-10-31
Fix the systemd unit path for fmn-backend

jeremycline:
It should be /usr/share, not /usr/lib/share


pypingou:
👍 for me

PR #248, opened 2017-11-01
Skip queries for watch flag rather than crashing

jeremycline:
Plenty of rules try to query on watch, so just skip them rather than
raising an exception when they query on watch.

CC @ralphbean 


pypingou:
Looks fine to me :)

PR #252, opened 2017-11-01
Set the routing key for broadcast messages

jeremycline:
We need to make sure broadcasts are routed to all workers, not just one out of many (otherwise only one worker refreshes its cache and the rest use stale caches).


pypingou:
👍 for me

PR #253, opened 2017-11-09
Announce message notifications via fedmsg

jeremycline:
Send a notification of the notification we're about to send via fedmsg.
Make sure the consumer does not try to process that notification.

@abompard I thought I'd just put this up and we could talk about the details of what you need on the PR. I was thinking about implementing a simple formatter for the message so it would arrive on your end as human-readable, but I thought you might want to control that so maybe it makes sense to just resend the original fedmsg. Also, I don't know whether the message topic should be "notification.{user}.{context}" or "notification.{context}.{user}" - the first lets you subscribe to all contexts for a user, or all contexts for all users, while the second lets you subscribe to all users for a context, or all contexts for all users.

### DIFF
--- a/fmn/consumer.py
+++ b/fmn/consumer.py
@@ -104,6 +104,11 @@ class FMNConsumer(fedmsg.consumers.FedmsgConsumer):
         """
         topic, msg = raw_msg['topic'], raw_msg['body']
 
+        if 'fmn.notification' in topic:
+            # This is a notification of a notification we've sent. Processing it
+            # would lead to a positive feedback loop.
+            return
+
         for suffix in self.junk_suffixes:
             if topic.endswith(suffix):
                 log.debug("Dropping %r", topic)
--- a/fmn/tasks.py
+++ b/fmn/tasks.py
@@ -79,6 +79,11 @@ class _FindRecipients(task.Task):
         """
         _log.info('Initializing the "%s" task', self.name)
         fedmsg.meta.make_processors(**config.app_conf)
+        try:
+            fedmsg.init()
+        except ValueError:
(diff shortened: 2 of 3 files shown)


abompard:
For Hubs I'll just listen to all context for all users, I think, so it doesn't matter. However outside of Hubs it feels like the most common use case would be to subscribe to all contexts for a user, so `{user}.{context}` feels like a better approach.

Do you have examples of messages with and without the simple formatter?

jeremycline:
The un-formatted message would just be a fedmsg with ``{'original_message': whatever_the_original_fedmsg_was}``. The formatted message would be that, with the message run through ``fedmsg.meta.msg2title`` and ``fedmsg.meta.msg2subtitle``. You'd also be able to do that on the Hubs end and it would let you easily adjust the format if necessary.

abompard:
Let's keep the original message then, thanks!

fmn/consumer.py
```diff
@@ -104,6 +104,11 @@ def work(self, session, raw_msg):
         """
         topic, msg = raw_msg['topic'], raw_msg['body']
 
+        if 'fmn.notification' in topic:
+            # This is a notification of a notification we've sent. Processing it
+            # would lead to a positive feedback loop.
+            return
```
abompard, line 110: Is this unit-tested? I think it should be, it's pretty important.

jeremycline: Definitely agree. Sorry for the slow response, I got caught up in a bunch of other stuff. After I get the next round of FMN bugfixes in and deployed, I'll write tests for this and get it in 2.1.

PR #254, opened 2017-11-16
Stop converting database models to dicts

jeremycline:
This should _vastly_ improve the performance of load_preferences. It
does this by 1) eagerly loading the filter and rule relationships (which
we need in all cases) and 2) keeping the preferences as the SQLAlchemy
database classes rather than turning them all to JSON.


bowlofeggs (approved):
Looks OK to me, but it would be good to make sure that we have test coverage on this patch.

jeremycline:
I had put this in production when fmn-2.0.0 was having lots of problems, but this patch introduced more problems than it solved. I'm going to just close this and we'll need to figure something else out at some point.

PR #255, opened 2017-11-16
Try to speed up getting the user->package mapping

jeremycline:
On the first message, we (unfortunately) have to get the packages a user owns/co-maintains/watches for every single user in FMN. This is because there are rules like "Notify me when X happens to a package I own/co-maintain/watch" and those rules are in the default rule set (so my guess is nearly everyone has them on. This means that when processing the first message, we block for a _really_ long time while those queries happen and get cached.

This ups the items per page to 100 (the max) and uses a session so we at least reuse the connection for each page. A more global session should be used, but care needs to be taken with that.

### DIFF
--- a/fmn/rules/utils.py
+++ b/fmn/rules/utils.py
@@ -31,6 +31,8 @@ _FAS = None
 # the fast part at the tightest part of the loop.
 _regex_cache = {}
 
+requests_session = requests.Session()
+
 
 def compile_regex(pattern):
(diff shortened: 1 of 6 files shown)


pypingou:
Looks fine to me, should we remove the forks while at it?

jeremycline:
@pypingou I've written the patch on top of this, I wanted to separate the two out so I'll just merge this and open a second PR.

PR #256, opened 2017-11-16
Run using prefork with concurrency 1

jeremycline:
Rather than running in solo mode, run in prefork mode with concurrency
of 1. The difference is long-running tasks (of which FMN unfortunately
has many) block the heartbeat celery uses to see if the workers are
alive. The reason we want concurrency=1 is that we use broadcasts to
tell workers to refresh their caches and if we don't do this, only one
process in the worker's pool will handle the message.


pypingou:
👍 for me

PR #258, opened 2017-11-17
Don't explode when logging the success of a batch delivery

jeremycline:
The fedmsg argument can be a list if the message was a batch, so don't
use a dict API. Prior to this commit, when a batch was delivered the log
message would raise an exception inside the try block which resulted in
the message being requeued for delivery (starting the process all over
again).


pypingou:
👍 looks like the error that got reported by email

PR #259, opened 2017-11-17
Ensure formatting doesn't result in an unhandled error

jeremycline:
When we fall back to sending the raw fedmsgs, we still need the email to
have sane headers. Ideally all formatters should ensure no exception is
raised and they always return _something_ sendable. This just converts
the email formatters, however. The rest can be handled in later commits.

### DIFF
--- a/fmn/formatters.py
+++ b/fmn/formatters.py
@@ -31,7 +31,6 @@ import datetime
 import time
 import email as email_module
 
-from kitchen.text.converters import to_bytes, to_unicode
 import arrow
 import requests
 import six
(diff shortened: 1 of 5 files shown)


fmn/formatters.py
```diff
@@ -177,7 +176,45 @@ def irc_batch(messages, recipient):
...
+            # This apparently returns a list of dicts with the 'link', 'subtitle',
+            # 'subjective', 'secondary_icon' and any number of other keys from the
+            # individual conglomerator.
+            conglomerated = fedmsg.meta.conglomerate(messages, **config.app_conf)
+            for message in conglomerated:
+                formatted_message += template.format(
+                    subtitle=message['subtitle'], delta=message.get('human_time', ''),
+                    link=message['link'], filter_link=_irc_filter_link(recipient))
+        except Exception:
+            formatted_message = (
```
puiterwijk: I am not sure we should pass this directly to users... It's not like they are able to fix things.
Also, regardless of whether you send this to users, you really should also log them, preferably including traceback, so we can actually figure out what went wrong...

fmn/formatters.py
```diff
@@ -403,78 +415,76 @@ def email_batch(messages, recipient):
...
-        for msg in messages])
-    subject = u'Fedora Notifications Digest ({n} updates)'.format(n=len(messages))
-    summary = u"Digest summary:\n"
-    i = 0
     for msg in messages:
-        i = i + 1
-        line = fedmsg.meta.msg2subtitle(msg, **config.app_conf) or u''
-        summary += '%d.\t%s\n' % (i, line)
+        try:
+            for username in fedmsg.meta.msg2usernames(msg, **config.app_conf) or []:
```
cverna: Line 369 we have the same block maybe this could be put inside a function to avoid duplication

fmn/formatters.py
```diff
@@ -403,78 +415,76 @@ def email_batch(messages, recipient):
...
+                email_message.add_header('X-Fedmsg-Username', username.encode('utf-8'))
+        except Exception:
+            _log.exception('fedmsg.meta.msg2usernames failed to handle %r', msg)
+        try:
+            for package in fedmsg.meta.msg2packages(msg, **config.app_conf) or []:
```
cverna: Same here with line 376.

cverna (approved):
Just a suggestion, but LGTM 👍

puiterwijk (changes requested):
Error handling in IRC case really should be fixed.

fmn/formatters.py
jeremycline: Nice catch, logging statement added.

It's true there's not much they can do about it, but it's also going to be a weird set of messages and I just wanted them to know it's not normal. I'd welcome better phrasing, though.

fmn/formatters.py
jeremycline: Ah, I see what you mean. I think I can factor out a lot of code with some tweaks to the _base_email() function.

fmn/formatters.py
puiterwijk: Maybe add a phrase like "This error has also been reported to the admins, but you can join #fedora-admin for more information"?
At least give users an idea of what to do with the info :).

fmn/formatters.py
```diff
@@ -178,7 +176,48 @@ def irc_batch(messages, recipient):
...
+            # This apparently returns a list of dicts with the 'link', 'subtitle',
+            # 'subjective', 'secondary_icon' and any number of other keys from the
+            # individual conglomerator.
+            conglomerated = fedmsg.meta.conglomerate(messages, **config.app_conf)
+            for message in conglomerated:
+                formatted_message += template.format(
+                    subtitle=message['subtitle'], delta=message.get('human_time', ''),
+                    link=message['link'], filter_link=_irc_filter_link(recipient))
+        except Exception:
+            _log.exception('An unexpected error occurred while processing an IRC batch')
```
puiterwijk, line 191: Just to be nitpicky, but maybe also log the message IDs in the batch here?

puiterwijk (approved):

fmn/formatters.py
jeremycline: Done

PR #260, opened 2017-11-18
Encode the email with UTF-8

jeremycline:
When formatting the email, we mark it as UTF-8. However, when it arrives
in the delivery service it is decoded to unicode until it's being sent.
This encodes it just before handing it to Twisted for delivery.


pypingou:
Looks fine to me

PR #264, opened 2017-11-30
Ignore forks when seeing what projects a user has

### LINKED ISSUE #261: Users are receiving notifications for repos they've merely forked
There are reports of users receiving notifications for packages they don't own or co-maintain, but have simply forked, similar to https://pagure.io/fedora-infrastructure/issue/6445.

This should be investigated since it's likely the new pagure rules causing this.

jeremycline:
Fixes #261


pypingou:
👍 for me

PR #265, opened 2017-11-30
Catch general SMTPClientError for Twisted

### LINKED ISSUE #262: Email messages are occasionally sent using <openid>@fedoraproject.org which is invalid.
For example, jcline.id.fedoraproject.org@fedoraproject.org rather than jcline@fedoraproject.org.

jeremycline:
When mail fails to be delivered with SMTP 550, Twisted raises a
SMTPClientError rather than SMTPBadRcpt. It turns out SMTPBadRcpt is for
the server side of things and I didn't read enough documentation.

Fixes #262


cverna (approved):
👍

PR #269, opened 2017-12-13
taskotron: remove upgradepath from critical tasks

kparal:
Upgradepath check has been deprecated. See
https://pagure.io/taskotron/issue/244


kparal:
Please note, I'm not really familiar with the project. If some database migration scripts or anything similar needs to be written as well, somebody else needs to step in.

jeremycline (approved):
Looks good, thanks!

PR #270, opened 2017-12-13
Configure the celery workers with the FMN logging config

### LINKED ISSUE #267: Logging in the fmn workers doesn't get configured correctly
Logging in the fmn workers doesn't get configured to send emails for errors/exception level. We need to use celery to use our fmn.config logging dict rather than its own.

jeremycline:
fixes #267


pypingou:
Looks fine to me

PR #272, opened 2017-12-21
Add SQLAlchemy error handling to tasks and the irc backend

jeremycline:


fmn/delivery/backends/irc.py
```diff
@@ -206,66 +216,71 @@ def subcmd_filters(self, nick, message):
         log.info("CMD list filters:  %r sent us %r" % (nick, message))
 
         valid_paths = fmn.lib.load_rules(root="fmn.rules")
-        sess = fmn.lib.models.init(self.config.get('fmn.sqlalchemy.uri'))
-        pref = self.get_preference(sess, nick)
+        sess = fmn.lib.models.Session.remove()
```
puiterwijk: Shouldn't this be `.Session()`?

jeremycline: No, ``fmn.lib.models.Session`` is a [scoped session](https://docs.sqlalchemy.org/en/latest/orm/contextual.html#sqlalchemy.orm.scoping.scoped_session.remove).

puiterwijk: Sure. But at this moment you are trying to get a new session, correct? Not closing this one, right at the start of the call?

puiterwijk: Specifically: you're assigned the result to the `Session.remove()` call to something, but that function doesn't even return anything? https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/orm/scoping.py#L76

puiterwijk (changes requested):
Please fix `sess = ..Session.remove()`.
Otherwise looks fine.

puiterwijk (approved):

PR #273, opened 2018-01-19
Only list at most 10 packages in email headers

puiterwijk:
Sometimes we get fedmsg's with a lot of packages mentioned[1].
This does not all fit into the email headers of a single email, so let's just
send the first 10 and a header indicating how many packages there were in
total.

1: https://apps.fedoraproject.org/datagrepper/id?id=2018-09f9d5d1-1f53-4488-9bfa-c08828a490b5&is_raw=true&size=extra-large

Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>


fmn/formatters.py
```diff
@@ -552,8 +553,13 @@ def _base_email(recipient=None, messages=None):
...
             try:
                 for package in fedmsg.meta.msg2packages(msg, **config.app_conf) or []:
-                    email_message.add_header('X-Fedmsg-Package', package.encode('utf-8'))
+                    packages.add(package)
             except Exception:
                 _log.exception('fedmsg.meta.msg2packages failed to handle %r', msg)
 
+        for pkg in list(packages)[:10]:
+            email_message.add_header('X-Fedmsg-Package', pkg.encode('utf-8'))
+        if len(packages) > 10:
```
pypingou: I'd even always include this one, then we're consistent.

puiterwijk: Done, thanks for the hint!

pypingou:
Travis seems unhappy but code looks fine to me

PR #280, opened 2018-03-09
Lowering digest content limits

mattiaverga:
This PR lowers the maximum number of characters admitted in the digest content from 2M to 500k and fixes a missing unicode declaration to the line where msg_ids are added to the content.

The PR also add a limit to the number of messages that can be included in a digest (1000). Now the user can set in backend a maximum number of 250 before get the digest, but it can also set that to 'disabled' and only put a temporal limit. If something goes wrong, a digest can include thousands of messages, so this limit will ensure a protection against that.
I've added a test unit to check this limit is working as expected.

p.s. I removed the second check on the digest content length because in the worst case we now have a content made by 999 msg_ids, so it's not possible that it overcomes the 500k chars limit.

### DIFF
--- a/fmn/formatters.py
+++ b/fmn/formatters.py
@@ -436,6 +436,14 @@ def email_batch(messages, recipient):
         return email(messages[0], recipient)
 
     email_message = _base_email(recipient=recipient, messages=messages)
+
+    if len(messages) >= 1000:
+        email_message.add_header('Subject', u'Fedora Notifications Digest error')
+        digest_error = (u'Too many messages were queued to be sent in this digest ({n})!\n'
+                        u'Consider adjusting your FMN settings.\n'.format(n=len(messages)))
+        email_message.set_payload(digest_error.encode('utf-8'), 'utf-8')
+        return email_message.as_string()
+
     email_message.add_header(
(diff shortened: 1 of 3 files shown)


mattiaverga:
mmmm... test is failing because of email headers.
Is the `X-Fedmsg-Username` supposed to add an entry for every message even if username is the same, or it should add a single entry for each user? If the latter, then `_base_email()` needs some tweaking.

mattiaverga:
I will make some more changes once the other PR is merged.

fmn/formatters.py
```diff
@@ -494,18 +502,12 @@ def email_batch(messages, recipient):
...
                           u'The following messages were batched:\n\n')
         for msg in messages:
-            digest_content += msg['msg_id'] + '\n'
-
-        if len(digest_content) > 20000000:
-            # Even the briefest summary is too big
-            digest_content = (u'The message digest was so large, '
-                              u'not even a summary could be sent.\n'
-                              u'Consider adjusting your FMN settings.\n')
+            digest_content += msg['msg_id'] + u'\n'
```
jeremycline, line 510: Nice simplification :+1:

jeremycline (approved):
Looks good, thanks! :shipit:

PR #281, opened 2018-03-09
X-Fedmsg-Username headers should not be duplicated.

mattiaverga:
As pointed out in #280 I see no reason to have multiple `X-Fedmsg-Username` with the same username value.
I think there must be only one unique `X-Fedmsg-Username` per username in email headers, so I applied the same method used for `X-Fedmsg-Package` and I added a test for it.


jeremycline (approved):
Looks great, thanks!

PR #283, opened 2018-03-11
Do not add X-Fedmsg headers when messages are too many

mattiaverga:
Do not add X-Fedmsg headers when messages are too many or we will have bad email.


jeremycline (approved):
Makes sense. Thanks!

PR #287, opened 2018-03-27
Add timestamp to single message emails

### LINKED ISSUE #285: [RFE] Please add the date of notification on the notification
I have a suggestion. After some confusions that I have made related with notifications with a big delay .

Please add the date of notification on the notification .

As we expect notifications on time , when don't have it and we receive an notification very delayed , it would help have the date of notification in notification .

mattiaverga:
Fixes #285 
This will add a timestamp as starting line for every single message email.
The other content (message or dump) remains unchanged.

### DIFF
--- a/fmn/formatters.py
+++ b/fmn/formatters.py
@@ -399,8 +399,9 @@ def email(message, recipient):
             subject_prefix.strip(), subject.strip())
     email_message.add_header('Subject', subject.encode('utf-8'))
 
-    timestamp = datetime.datetime.fromtimestamp(message['timestamp'])
-    content = timestamp.strftime('%c') + u'\n'
+    timestamp = datetime.datetime.fromtimestamp(message['timestamp'], tz=pytz.utc)
+    timestamp = timestamp.strftime('%Y-%m-%d %H:%M:%S %Z')
+    content = u'Notification time stamped {}\n\n'.format(timestamp)
     try:
         content += fedmsg.meta.msg2long_form(message, **config.app_conf) or u''
     except Exception:
@@ -475,10 +476,10 @@ def email_batch(messages, recipient):
                 _log.exception('fedmsg.meta.msg2link failed to handle %r', message)
                 link = u'No link could be found in the message'
 
-            timestamp = datetime.datetime.fromtimestamp(message['timestamp'])
+            timestamp = datetime.datetime.fromtimestamp(message['timestamp'], tz=pytz.utc)
             summary.append(summary_template.format(number=message_number, short_title=shortform))
             formatted_messages.append(message_template.format(
-                ts=timestamp.strftime('%c'), short_title=shortform,
+                ts=timestamp.strftime('%Y-%m-%d %H:%M:%S %Z'), short_title=shortform,
                 link=link,
                 details=longform))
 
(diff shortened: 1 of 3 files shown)


fmn/formatters.py
```diff
@@ -399,11 +399,13 @@ def email(message, recipient):
             subject_prefix.strip(), subject.strip())
     email_message.add_header('Subject', subject.encode('utf-8'))
 
+    timestamp = datetime.datetime.fromtimestamp(message['timestamp'])
+    content = timestamp.strftime('%c') + u'\n'
```
jeremycline: It's not documented anywhere AFAIK, but those timestamps are UTC so we should note that to avoid confusion. Maybe something like:
```
import pytz
utc_timestamp = pytz.utc.localize(message['timestamp']).strftime('%Y-%m-%d %H:%M:%S %Z')
content = u'Notification time stamped {}\n\n'.format(utc_timestamp)
```
Sorry, I know this is going to make you edit all those test results again :(  - I should have been less lazy and added something to encode the expected results with base64 automatically.

jeremycline (changes requested):
Thanks so much for handling this, there's just one tweak needed to avoid confusion down the road

fmn/formatters.py
mattiaverga: It would make more sense if the timezone is already set in the received message['timestamp'] (is that from fedmsg?).
Maybe we can ask fedmsg to implement that, in the meantime I can just modify here to set the timezone if the received message['timestamp'] is not already timezone aware.

And I think I should also modify the timestamps in digest creation in the same way.
What do you think?

jeremycline: > It would make more sense if the timezone is already set in the received message['timestamp'] (is that from fedmsg?) Maybe we can ask fedmsg to implement that, in the meantime I can just modify here to set the timezone if the received message['timestamp'] is not already timezone aware.

So looking at the fedmsg source, it's just ``time.time()`` which is seconds since the epoch (January 1, 1970, 00:00:00 (UTC)): https://github.com/fedora-infra/fedmsg/blob/a0547bb08372ade44adb088bc57de34f0f86e729/fedmsg/core.py#L290. This makes it easy for consumers to convert it to whatever timezone they desire, so I think it's best to leave that as it is.

I did realize my suggestion is wrong, though, since pytz.utc.localize wants datetime objects. An easier implementation is just ``datetime.datetime.fromtimestamp(message['timestamp'], tz=pytz.utc)``.

> And I think I should also modify the timestamps in digest creation in the same way.
What do you think?

Yes, that's a good idea - you should just need to add that tz kwarg.

mattiaverga: Yes, I was confusing timestamps with datetime objects too... :-(
I'll work on it in the next days as soon as I have some free time.

jeremycline: Cool, thanks!

mattiaverga:
It seems that the failure is not due to the PR, but to an internal error of Travis, doesn't it?

jeremycline (approved):

jeremycline:
Yeah, Travis is occasionally flaky. What I tend to do is just ``git commit --amend`` and force push to retrigger the tests. Thanks for this!

PR #292, opened 2018-04-03
Use single quotes in SQL for the git.repo.new migration

jeremycline:
Single quotes and double quotes are not interchangeable, it seems. It
also seems that SQLite is fine with double quotes, but Postgres is
*not*.


pypingou (approved):
Ran into this one as recently as 1h ago...

Issue #100, opened 2015-07-09
HTML email marked as text/plain


mizdebsk:
In some cases FMN can send HTML email with `Content-Type: text/plain`. Resulting message is unreadable.

Reproducer: push code to newly created git branch (less than one minute after branch creation).

Example raw message (with irrelevant headers stripped):

```
Return-Path: notifications@fedoraproject.org
To: mizdebsk@fedoraproject.org
From: notifications@fedoraproject.org
X-Fedmsg-Topic: org.fedoraproject.prod.git.receive
X-Fedmsg-Category: git
X-Fedmsg-Username: mizdebsk
X-Fedmsg-Package: pmd-build-tools
Subject: mizdebsk pushed to pmd-build-tools (master).  "Upload sources"
MIME-Version: 1.0
Content-Type: text/plain; charset="utf-8"
Content-Transfer-Encoding: base64
Message-Id: <20150518201027.3FFFD605DFD5@bastion01.phx2.fedoraproject.org>
Date: Mon, 18 May 2015 20:10:27 +0000 (UTC)

PCFET0NUWVBFIGh0bWwgUFVCTElDICItLy9XM0MvL0RURCBYSFRNTCAxLjAgVHJhbnNpdGlvbmFs
Ly9FTiIKICAiaHR0cDovL3d3dy53My5vcmcvVFIveGh0bWwxL0RURC94aHRtbDEtdHJhbnNpdGlv
bmFsLmR0ZCI+CjxodG1sIHhtbG5zPSdodHRwOi8vd3d3LnczLm9yZy8xOTk5L3hodG1sJyB4bWw6
bGFuZz0nZW4nIGxhbmc9J2VuJz4KPGhlYWQ+Cjx0aXRsZT5GZWRvcmEgUHJvamVjdCBQYWNrYWdl
cyBHSVQgcmVwb3NpdG9yaWVzPC90aXRsZT4KPG1ldGEgbmFtZT0nZ2VuZXJhdG9yJyBjb250ZW50
PSdjZ2l0IHYwLjEwLjInLz4KPG1ldGEgbmFtZT0ncm9ib3RzJyBjb250ZW50PSdpbmRleCwgbm9m
b2xsb3cnLz4KPGxpbmsgcmVsPSdzdHlsZXNoZWV0JyB0eXBlPSd0ZXh0L2NzcycgaHJlZj0nL2Nn
[...]

Issue #95, opened 2015-05-06
Sign emails with gpg


ralphbean:
People could verify that notifications really came from Fedora Infrastructure.

It would be cool to have more signed email accross the net, more desire for people to verify them, etc.

ryanlerch:
The FMN rewrite is almost complete. This issue was reported against the old version of FMN, and likely does not apply anymore. 

Closing, but reopen if you feel this still applies.

PR #294, opened 2018-04-22
Refactor mail creation using MIME classes

### LINKED ISSUE #92: Make digests be actual digests
The "digests" sent by FMN don't appear to be actual email digests.  As such, mail readers which can unpack digests don't unpack them properly.  Please consider using either standard MIME digests, or RFC 1153-format digests.

### LINKED ISSUE #101: Add a fedmsg message id header to emails
See #100.  There are `X-Fedmsg-...` headers, but it would be nice to know _exactly_ which fedmsg triggered that email.  The original messages have a unique id which can be used to look them up in datagrepper.

### LINKED ISSUE #133: Missing i18n in subjects of the sent emails from notifications@fedoraproject.org
Hi,
I'm receiving emails from notifications@fedoraproject.org without i18n in the subjects

The subject sent from notifications@fedoraproject.org was:
fredlima posted "V??deo sobre a hist??ria da antiga distribui????o Red Hat Linux at?? o RHEL"

but it will be correct if it was sent like this:
fredlima posted "Vídeo sobre a história da antiga distribuição Red Hat Linux até o RHEL"

the message body text is correct.

can you take a look at this issue?
thank you in advance

mattiaverga:
I've modified the way email messages are created by using MIME classes provided by the email module.

The main purpose was to make it more compatible with Python 3 and also to create standard compliant digests, so that it fixes #92 

With this change a digest (in verbose mode) is made up of a multipart email container, with a text/plain message summary and the real email messages attached as message/rfc822 messages. The non verbose digest is a simple text/plain email, just like before.
This may change significantly how the digest is displayed to user. With thunderbird there is not so much difference, because attached messages are displayed after the summary, but in webmail it may be possible that only summary is displayed and full messages are available only as attachments.
As such, I would like some other users to test this change.

I've also added fedmsg msg_id to headers (fixes #101 ) and changed how headers are added to correctly support i18n (fixes #133 ).

Using MIME classes opens up the way to support signed emails (#95 ) and html emails (#100 ), but I did not start working on those yet. Is that something you are interested in having? I'm particularly doubtful especially on html emails (is there anything that still produces fmn messages with email content? the example in the reported bug would not display correctly even within a text/HTML message...)

### DIFF
--- a/fmn/formatters.py
+++ b/fmn/formatters.py
@@ -29,7 +29,11 @@ import json
 import uuid
 import datetime
 import time
-import email as email_module
+from email import message_from_string
+from email.mime.nonmultipart import MIMENonMultipart
+from email.mime.multipart import MIMEMultipart
+from email.mime.message import MIMEMessage
+from sys import getsizeof
 
 import arrow
 import requests
@@ -386,7 +390,9 @@ def email(message, recipient):
     Returns:
         str: The email as a unicode string.
     """
-    email_message = _base_email(recipient=recipient, messages=[message])
+    email_message = _base_email(MIMENonMultipart('text', 'plain'),
+                                recipient=recipient,
+                                messages=[message])
 
     try:
         subject = fedmsg.meta.msg2subtitle(message, **config.app_conf) or u'fedmsg notification'
@@ -432,10 +438,6 @@ def email(message, recipient):
         content += u'\n\n--\n{0}'.format(footer.strip())
     email_message.set_payload(content.encode('utf-8'), 'utf-8')
 
-    # Explicitly declare encoding, but remove the transfer encoding
-    # https://github.com/fedora-infra/fmn/issues/94
-    email_message.set_charset('utf-8')
-
     return email_message.as_string()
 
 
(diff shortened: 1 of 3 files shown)


jeremycline:
Hey, thanks for the PR. I just wanted to let you know I'm not ignoring this, I've just got a bunch of other stuff on my plate at the moment. I'll see about reviewing it early next week if possible.

mattiaverga:
Sure, take your time! :-)

fmn/formatters.py
```diff
@@ -548,43 +551,50 @@ def email_confirmation(confirmation):
...
             for cat in categories:
-                email_message.add_header('X-Fedmsg-Category', cat.encode('utf-8'))
+                email_message.add_header('X-Fedmsg-Category', cat)
         except Exception:
             _log.exception('Unable to parse message topic and category for %r', messages)
 
         usernames = set()
         packages = set()
         for msg in messages:
+            email_message.add_header('X-Fedmsg-Id', msg['msg_id'])
```
jeremycline, line 595: This could be a maximum of 1000 headers, right? We had an issue with too many headers before (#273), maybe @puiterwijk  knows what an acceptable limit is? The package header was limited to 10, so I'm guessing it's less than 1000.

jeremycline (commented):
Hey, sorry this took so long to review. I have one comment which could do with input from @puiterwijk.

Overall, it looks like a great improvement. I'd like to hear from Patrick and test this locally (I'm on PTO through next week so I may get to it, I might not) before I approve it, though.

My only other request is that, if possible, you could rebase and squash your commits so this PR is a 3 commit series with the fixes for #92, #101, and #133 are each in their own commit. That way it's easy to revert individual fixes if one gives us trouble in staging.

Also, @abompard is going to be taking over the maintenance of fmn so he should also probably give this a look over.

Thanks again!

fmn/formatters.py
mattiaverga: Sorry, somehow Github stopped sending me emails, so I've just see your review...
I will rebase the commits and give a look to the maximum headers number ASAP.

Thanks

mattiaverga: About X-Fedmsg-Id header: if the recipient is set to 'verbose', the new digest is made of many email messages, each one with its own header ID.

The only problem is when the recipient is set to 'not verbose', in this case the batch_email is not a real digest, but a self made summary like the old one. So in this case we can have a single email with up to 1000 headers ID.

I can't find any limit to the number of headers an email can carry in the standards. Also I don't think the solution applied to the package header would be appropriate in this case. I'm open to suggestions.

puiterwijk: That limit is a header size limit, which is configured on pretty much all MTAs.
A concrete example: http://www.postfix.org/postconf.5.html, search for `header_size_limit (default: 102400)`.
Technically we could increase it for our mail server, but that only increases the chance that it gets refused by the recipients mailserver.

puiterwijk: An alternative idea might be to just store the message IDs for the batched message in the body itself..

mattiaverga: If so, we have troubles with the other X-Fedmsg headers as well. Each message adds:
- one X-Fedmsg-Topic
- one or more X-Fedmsg-Category
- one X-Fedmsg-Id
- one or more (up to ten) X-Fedmsg-Package
- one X-Fedmsg-Num-Packages
- one or more X-Fedmsg-Username

With an `header_size_limit` of 102400 even limiting the recap to 250 messages (the maximum that can be set on the web UI) can easily overwhelm the headers limit.

This can only impact on the 'recap' email, not on the real digest email, because in the digest the X-Fedmsg headers are in the attached messages.
So I would propose to not add the X-Fedmsg headers in the recap at all... but, since I don't know exactly the real word use case of those headers, I probably be wrong and break some other automated tool. Note that also setting the X-Fedmsg headers inside the attached messages instead of setting them in the top level digest can break those tools.

jeremycline: Sorry, I let this slip through the cracks. I *think* I'd be okay without including the X-Fedmsg headers in recaps. They make some amount of sense for filtering individual emails, but probably not with a huge digest.

pypingou:
@mattiaverga it looks like this is now conflicting. Do we want to solve the conflict or drop the change?

mattiaverga:
:open_mouth: I've completely forgot of this... I've quickly rebased the changes, but I was unable to run the tests locally since the Vagrant box is still based on F27. Let's see how Travis tests gone, if they fail I'll need some time to refresh my memory.

Zlopez:
The vagrant was now [bumped to F34](https://github.com/fedora-infra/fmn/pull/328).

Zlopez:
recheck

nphilipp:
@mattiaverga @Zlopez As we're working on a rewrite of FMN, should this PR be closed?

mattiaverga:
> @mattiaverga @Zlopez As we're working on a rewrite of FMN, should this PR be closed?

Yep, feel free to close it if you think it's too much effort to rebase / doesn't worth the effort.

nphilipp:
Yeah, we have a completely different code base i.e. also have to implement email related functionality anew. Thanks nonetheless!

PR #296, opened 2018-05-13
Don’t suggest to clone fmn.web and fmn.sse repositories

mavit:
The READMEs for each of these repositories says:

> This repository is kept for historical purposes only. It was merged into the fmn repository and you should look there if you would like to contribute or report an issue!


jeremycline (approved):
Thanks! :shipit:

PR #310, opened 2020-06-09
add ability to interact with fasjson

StephenCoady:
This commit adds the ability to interact with the new fasjson API. It builds its cache depending on the `fasjson` config flag from either FAS or the new fasjson.

### DIFF
--- a/fmn/config.py
+++ b/fmn/config.py
@@ -444,6 +444,13 @@ class _FmnConfig(dict):
             'default': {'username': None, 'password': None},
             'validator': dict,
         },
+        'fasjson': {
+            'default': {
+                'active': False,
+                'url': None
+            },
+            'validator': dict,
+        },
         'fmn.web.default_login': {
             'default': 'login',
             'validator': None,
--- a/fmn/consumer.py
+++ b/fmn/consumer.py
@@ -25,6 +25,7 @@ from .util import (
     new_packager,
     new_badges_user,
     get_fas_email,
+    get_fasjson_email
 )
 from fmn.tasks import find_recipients, REFRESH_CACHE_TOPIC, heat_fas_cache
 
@@ -163,7 +164,11 @@ class FMNConsumer(fedmsg.consumers.FedmsgConsumer):
                 log.info("Autocreating account for %r" % username)
                 openid = '%s.id.fedoraproject.org' % username
                 openid_url = 'https://%s.id.fedoraproject.org' % username
-                email = get_fas_email(config.app_conf, username)
+                fasjson = config.app_conf['fasjson'].get('active')
+                if fasjson:
+                    email = get_fasjson_email(config.app_conf, username)
+                else:
+                    email = get_fas_email(config.app_conf, username)
                 user = fmn.lib.models.User.get_or_create(
                     session, openid=openid, openid_url=openid_url,
                     create_defaults=True, detail_values=dict(email=email),
--- a/fmn/consumer.py
+++ b/fmn/consumer.py
@@ -164,7 +164,7 @@ class FMNConsumer(fedmsg.consumers.FedmsgConsumer):
                 log.info("Autocreating account for %r" % username)
                 openid = '%s.id.fedoraproject.org' % username
                 openid_url = 'https://%s.id.fedoraproject.org' % username
-                fasjson = config.app_conf['fasjson'].get('active')
+                fasjson = config.app_conf.get("fasjson", {}).get("active")
                 if fasjson:
                     email = get_fasjson_email(config.app_conf, username)
                 else:
--- a/fmn/fasjson_client.py
+++ b/fmn/fasjson_client.py
@@ -1,0 +1,55 @@
+import requests
+import requests.exceptions
+from requests.compat import urlencode, urljoin
+from requests_gssapi import HTTPSPNEGOAuth
+
+
+class Client(object):
+    """
+    A fasjson client to make very specific requests to fasjson.
+    Necessary because the official fasjson-client library does not support
+    python3.
+    """
+    def __init__(self, url, principal=None):
+        self.url = url
+        self.principal = principal
+
+        gssapi_auth = HTTPSPNEGOAuth(
+            opportunistic_auth=True, mutual_authentication="OPTIONAL"
+        )
+        self.session = requests.Session()
+        self.session.auth = gssapi_auth
+
+    def search(self, email):
+        """
+        A very limited search built to only serve fmn's requirement of
+        finding a user based on an email.
+        """
+        # email must be an exact match in fasjson, so we will either have
+        # 1 result or empty result
+        search_string = "search/users" + "?" + urlencode({"email": email})
+        endpoint = urljoin(self.url, search_string)
+
+        return self.session.get(endpoint)
+
+    def get_user(self, username):
+        """
+        Get a specific user based on their username
+        """
+        url_string = "users/" + username + "/"
+        endpoint = urljoin(self.url, url_string)
+
+        return self.session.get(endpoint)
+
+    def list_all_entities(self, ent_name):
+        """
+        Return all entities of a certain type. In fmn's case it is users.
+        """
+        endpoint = urljoin(self.url, ent_name + "/")
+
+        next_page_url = endpoint + "?" + urlencode({"page_number": 1})
+        while next_page_url:
+            res = self.session.get(next_page_url)
+            for item in res["result"]:
+                yield item
+            next_page_url = res.get("page", {}).get("next_page")
--- a/fmn/fasjson_client.py
+++ b/fmn/fasjson_client.py
@@ -1,22 +1,29 @@
+import logging
+
 import requests
 import requests.exceptions
+from gssapi import Credentials, exceptions
 from requests.compat import urlencode, urljoin
 from requests_gssapi import HTTPSPNEGOAuth
+
+
+log = logging.getLogger(__name__)
 
 
 class Client(object):
     """
     A fasjson client to make very specific requests to fasjson.
     Necessary because the official fasjson-client library does not support
-    python3.
+    python2.
     """
     def __init__(self, url, principal=None):
         self.url = url
         self.principal = principal
-
-        gssapi_auth = HTTPSPNEGOAuth(
-            opportunistic_auth=True, mutual_authentication="OPTIONAL"
-        )
+        try:
+            creds = Credentials(usage="initiate")
+        except exceptions.GSSError as e:
+            log.error("GSError. Unable to create credentials store.", e)
+        gssapi_auth = HTTPSPNEGOAuth(opportunistic_auth=True, creds=creds)
         self.session = requests.Session()
         self.session.auth = gssapi_auth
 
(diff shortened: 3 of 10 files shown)


fmn/fmn_fasshim.py
```diff
@@ -23,11 +24,34 @@
...
+    log.warn("Building the FASJSON cache into redis.")
+    if _cache.get('fas_cache_built'):
+        log.warn("FASJSON cache already built into redis.")
+        return
+    global client
+    try:
+        page_number = 0
+        next_page_exists = True
+        while next_page_exists:
+            page_number +=1
```
abompard: Flake8 will be unhappy here too. If this cache construct is frequent, maybe we could migrate it to `fasjson_client` itself, no?

abompard (changes requested):

fmn/fmn_fasshim.py
StephenCoady: I thought about that yeah - but I was unsure because I wasn't sure what the client's request would look like. Would it just be the current `get_users` but with pagination built in?

fmn/fmn_fasshim.py
```diff
@@ -57,79 +80,85 @@ def make_fas_cache(**config):
...
+def _add_to_cache(users):
+    for user in users:
+        nick = user.get('ircnick')
         if nick:
-            _cache.set(nick, user['username'])
+            _cache.set(str(nick), user['username'])
+        else:
+            nicks = user.get('ircnicks', [])
+            for nick in nicks:
+                _cache.set(str(nick), user['username'])
```
abompard: Is this not a string already?

fmn/fmn_fasshim.py
```diff
@@ -57,79 +80,85 @@ def make_fas_cache(**config):
...
-    for user in request['people']:
-        nick = user['ircnick']
+def _add_to_cache(users):
+    for user in users:
+        nick = user.get('ircnick')
         if nick:
-            _cache.set(nick, user['username'])
+            _cache.set(str(nick), user['username'])
+        else:
+            nicks = user.get('ircnicks', [])
```
abompard: Instead of dealing with the differences between the FAS and the FASJSON models here, I would recommend editing the results from FAS to set `user["ircnicks"] = [user["ircnick"]]`, same thing for email / emails.

fmn/fmn_fasshim.py
```diff
@@ -57,79 +80,85 @@ def make_fas_cache(**config):
...
+            for nick in nicks:
+                _cache.set(str(nick), user['username'])
 
-        email = user['email']
+        email = user.get('email')
         if email:
-            _cache.set(email, user['username'])
+            _cache.set(str(email), user['username'])
```
abompard: This too should already be a string

fmn/fmn_fasshim.py
```diff
@@ -57,79 +80,85 @@ def make_fas_cache(**config):
...
+        else:
+            emails = user.get('emails', [])
+            for email in emails:
+                _cache.set(str(email), user['username'])
```
abompard: same here

abompard (changes requested):

fmn/fmn_fasshim.py
abompard: This is still valid, the `nick` and `email` values should already be strings, no need to convert them. Or are you thinking of cases where it's not?

fmn/fmn_fasshim.py
```diff
@@ -57,79 +74,96 @@ def make_fas_cache(**config):
...
-        for user in request['people']:
-            nick = user['ircnick']
-            if nick:
-                _cache.set(str(nick), user['username'])
-
-            email = user['email']
-            if email:
-                _cache.set(str(email), user['username'])
-
+        _add_to_cache(request['people'])
```
abompard: I don't think you should do that replacement here, `_add_to_cache()` does not know how to handle the old fas2 format anymore.

fmn/fmn_fasshim.py
```diff
@@ -57,79 +74,96 @@ def make_fas_cache(**config):
...
-        if nick:
-            _cache.set(nick, user['username'])
-
-        email = user['email']
-        if email:
-            _cache.set(email, user['username'])
+    global client
+    if config.get('fasjson'):
+        try:
+            log.info("Downloading FASJSON cache for %s*" % username)
```
abompard: Flake8 caught that this should be `email` because `username` is undefined. Flake8 is nice ;-)

fmn/fmn_fasshim.py
```diff
@@ -23,11 +24,28 @@
...
+    log.warn("Building the FASJSON cache into redis.")
+    if _cache.get('fas_cache_built'):
+        log.warn("FASJSON cache already built into redis.")
+        return
+    global client
+    try:
+        _add_to_cache(list(client.list_all_entities("users")))
+    except fasjson_client.errors.APIError as e:
+        log.error("Something went wrong building cache with error: %s" % e)
+        continue
```
abompard: Flake8 also caught that you can't have a `continue` statement outside of a loop, it should be `return`.

abompard (changes requested):

fmn/config.py
```diff
@@ -444,6 +444,10 @@ class _FmnConfig(dict):
             'default': {'username': None, 'password': None},
             'validator': dict,
         },
+        'fasjson': {
+            'default': False,
+            'validator': bool,
+        }
```
abompard: Missing comma here

fmn/fmn_fasshim.py
StephenCoady: No, sorry. With all the different PRs and this one and fedmsg_meta being similar I just missed it! Thanks.

fmn/fmn_fasshim.py
StephenCoady: Same as before, missed this. Thanks.

fmn/util.py
```diff
@@ -36,3 +37,16 @@ def get_fas_email(config, username):
...
+
+
+def get_fasjson_email(config, username):
+    """ Return FASJSON email associated with a username. """
+    try:
+        fasjson = config["fasjson"]
+        client = fasjson_client.Client(url=fasjson.get('url'))
+        person = client.get_user(username=username).result
+
+        return person.get('emails')[0]
```
StephenCoady, line 49: not sure what the best way here is, fasjson will provide multiple emails and the data structure calling this functions expects one.

StephenCoady:
So I'm not sure what the best way forward here is. The CI is failing and to make it pass there is a lot of work.

The first blocker is removing python 2 support. I'm not sure of the implications of doing that. I think things like that might be best sorted in another PR. At the moment I can't run the tests locally, there are lots of errors even on py36.

fmn/fasjson_client.py
```diff
@@ -0,0 +1,55 @@
...
+import requests.exceptions
+from requests.compat import urlencode, urljoin
+from requests_gssapi import HTTPSPNEGOAuth
+
+
+class Client(object):
+    """
+    A fasjson client to make very specific requests to fasjson.
+    Necessary because the official fasjson-client library does not support
+    python3.
```
abompard: Typo, it does not support python**2**

fmn/fasjson_client.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    def __init__(self, url, principal=None):
+        self.url = url
+        self.principal = principal
+
+        gssapi_auth = HTTPSPNEGOAuth(
+            opportunistic_auth=True, mutual_authentication="OPTIONAL"
```
abompard: Looks like `mutual_authentication="OPTIONAL"` [is discouraged](https://github.com/pythongssapi/requests-gssapi#optional) by requests-gssapi. It should probably be disabled, if I understand the docs well.

abompard: In a [python-freeipa ticket](https://github.com/opennode/python-freeipa/issues/46), Crys recommends creating the credentials and attaching the object to the HTTPSNEGOAuth instance. I'm not entirely sure what the benefits are but I've found out it's always safer to follow Crys' advice ;-)

fmn/fmn_fasshim.py
```diff
@@ -71,65 +91,92 @@ def make_fas_cache(**config):
...
             _cache.set(email, user['username'])
+
+
+def update_nick(username):
+    global client
+    if config.get('fasjson'):
+        try:
+            log.info("Downloading FASJSON cache for %s*" % username)
+            response = client.get_user(username=username)
+            _add_to_cache([response.json().result])
```
abompard: Are you sure it's not `.json()["result"]` ? Also, I think you could move the `.json()` calls into the `Client` class.

fmn/util.py
abompard: Yeah getting the fist one seems a good default, even though the LDAP entries are not sorted (so the "first" in the list could be different between calls).

fmn/util.py
```diff
@@ -1,4 +1,5 @@
 import fedora.client
+import fasjson_client
```
abompard, line 2: You should import it as `from . import fasjson_client` or `from .fasjson_client import Client`, otherwise it'll use the system-wide `fasjson_client` if it is installed.

requirements.txt
```diff
@@ -6,6 +6,7 @@ celery>=4.0
 datanommer.models
 docutils
 dogpile.cache
+fasjson-client
```
abompard: You don't require it anymore I think.

abompard (changes requested):
I haven't tested it yet, here are a few comments / changes to make.

fmn/consumer.py
```diff
@@ -163,7 +164,11 @@ def work(self, session, raw_msg):
                 log.info("Autocreating account for %r" % username)
                 openid = '%s.id.fedoraproject.org' % username
                 openid_url = 'https://%s.id.fedoraproject.org' % username
-                email = get_fas_email(config.app_conf, username)
+                fasjson = config.app_conf['fasjson'].get('active')
```
pypingou: I don't know if we can assume that ``fasjson`` is always there, if we can't we could do something like ``app_conf.get("fas_json", {}).get("active")``

pypingou:
Still failing on ``requests_gssapi``

fmn/consumer.py
abompard: +1 to that

fmn/fmn_fasshim.py
```diff
@@ -23,27 +25,45 @@
...
+
+
+def make_fasjson_cache(**config):
+    log.warning("Building the FASJSON cache into redis.")
+    if _cache.get('fas_cache_built'):
+        log.warning("FASJSON cache already built into redis.")
+        return
+    global client
+    try:
+        _add_to_cache(list(client.list_all_entities("users")))
```
abompard, line 46: From a memory point of view it would be better to loop over `list_all_entities` than building the list in memory. Something like:
```
for users in client.list_all_entities("users"):
    _add_to_cache(users)
```

But that's not a blocker.

abompard (approved):
Yep, looks good. I commented on a possible minor improvement, but I think it's good to go as is.

PR #318, opened 2021-05-07
Drop python bunch for python munch (which is py3 compatible!)

pypingou:
Signed-off-by: Pierre-Yves Chibon <pingou@pingoured.fr>


AdamSaleh:
Looks very non-controversial :+1:

pypingou:
Thanks!

PR #319, opened 2021-05-07
Fix the MANIFEST file

pypingou:


AdamSaleh:
From discussion:

```
asaleh: to be honest, I just ran check-manifest and now I am pretending to understand stuff :D https://gist.github.com/AdamSaleh/2a6a397344d7f8fb5e06ce3030c77b37
pingou: we don't want ansible in the tarball and nor the tests either, otherwise they get installed :/
asaleh: yeah, I figured, that is why I only mentioned those docs
pingou: the missing pieces at the top are safe to ignore
```

taking that into account, LGTM.

PR #323, opened 2021-09-01
update the keytab location and fix config accessor

StephenCoady:
this should fix the issue on notifs02

### DIFF
--- a/fmn/fmn_fasshim.py
+++ b/fmn/fmn_fasshim.py
@@ -104,7 +104,7 @@ def _add_to_cache(users):
 
 def update_nick(username):
     global client
-    if config.get('fasjson'):
+    if fasjson.get('active'):
         try:
             log.info("Downloading FASJSON cache for %s*" % username)
             response = client.get_user(username=username)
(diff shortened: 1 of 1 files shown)


fmn/fasjson_client.py
```diff
@@ -20,6 +21,7 @@ def __init__(self, url, principal=None):
         self.url = url
         self.principal = principal
         creds = None
+        os.environ["KRB5_CLIENT_KTNAME"] = "/etc/krb5.keytab"
```
abompard: That should probably be set on the system in `/etc/krb5.conf.d`, in case sysadmins want to put the keytab in a different place (that's common when dealing with openshift pods for example, because configs are mounted). So I think this change should be in Ansible.

StephenCoady: thanks, addressed now. i'll make a seperate pr for that on ansible

abompard (approved):

StephenCoady:
@pypingou could you give me access to this repo, please?

PR #324, opened 2021-09-08
use datanommer depending on the config flag

StephenCoady:
currently the 'datanommer.enabled' flag is
defined in config but is used nowhere in the
code. this commit uses it again.

### DIFF
--- a/fmn/web/app.py
+++ b/fmn/web/app.py
@@ -569,14 +569,17 @@ def example_messages(openid, context, filter_id, page, endtime):
         ))
 
     try:
-        total, pages, messages = datanommer.models.Message.grep(
-            start=end-delta,
-            end=end,
-            rows_per_page=bazillion,
-            page=page,
-            order='desc',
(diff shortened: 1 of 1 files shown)


fmn/web/app.py
```diff
@@ -569,14 +569,16 @@ def example_messages(openid, context, filter_id, page, endtime):
...
+        if config.app_conf['datanommer.enabled']:
+            total, pages, messages = datanommer.models.Message.grep(
+                start=end-delta,
+                end=end,
+                rows_per_page=bazillion,
+                page=page,
+                order='desc',
+                **hinting
+            )
+        raise Exception("Datanommer is not enabled.")
```
abompard: Aren't you missing an `else:` here? Looks like the exception is raised unconditionaly.

StephenCoady: :man_facepalming:

abompard (approved):

PR #332, opened 2021-12-14
Remove fasjson_client module

Zlopez:
The fasjson_client module was created because fasjson-client library was
only python3 and FMN was only python 2 at the time. This is no longer
true, so this commit removes the fasjson_client.py from FMN.

This should help with https://pagure.io/fedora-infrastructure/issue/9746

Signed-off-by: Michal Konečný <mkonecny@redhat.com>


Zlopez:
recheck

Zlopez:
recheck

morucci:
recheck

Zlopez:
recheck

Zlopez:
recheck

PR #334, opened 2022-01-25
Add override option to zuul

Zlopez:
Signed-off-by: Michal Konečný <mkonecny@redhat.com>


Zlopez:
The CI is now running, although it's failing, but this is the issue that isn't with the CI itself.

PR #335, opened 2022-02-01
Fix CI tests

Zlopez:
This PR fixes CI failures that are currently happening.


Zlopez:
This issue is strange, I can't reproduce it locally at all. Even the patch error happening earlier wasn't happening on my machine.

Zlopez:
recheck

PR #343, opened 2022-04-12
remove datanommer from vagrant and use postgres

ryanlerch:
remove setting up datanommer in the vagrant setup (we disable it by
default now) and switch the postgres groundwork we were using for
datanommer to use it for the main fmn database (rather than using sqlite)


Zlopez (approved):
Good one

PR #418, opened 2022-08-09
Add linting check to the frontend

ryanlerch:
This PR adds a linting check for frontend code, and also runs eslint --fix over the code too.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -35,3 +35,24 @@ jobs:
         run: |
              source venv/bin/activate
              tox
+
+  frontend-lint:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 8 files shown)


.github/workflows/main.yaml
```diff
@@ -35,3 +35,24 @@ jobs:
         run: |
              source venv/bin/activate
              tox
+
+  frontend-lint:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
```
abompard, line 44: Do we need these 3 lines?

ryanlerch: the ci joib above did it like that, so i just kept it the same

not sure why the ci job does it that way though

abompard: OK cool, we'll clean it in one go if we need to.

abompard (approved):

PR #421, opened 2022-08-10
prototype connecting to FASJSON 

ryanlerch:
This adds the basic funtionality of connecting to FASJSON and returning
the information back to the frontend. It also sets up the vagrant setup
for this as well.


nphilipp (approved):
Looks good to me!

nphilipp:
Fixed import sort order.

abompard:
You need to `dnf install -y krb5-devel` in the github action, I believe.

abompard:
Now that I think of it, we may not be able to use the `fasjson_client` helper, because it's not async. We may have to do the HTTP queries by hand (with HTTPX maybe?)

PR #451, opened 2022-08-16
First initial early example of a "New Rule" page

abompard:


abompard:
@ryanlerch you can have a look at that if you want, but it's very basic

PR #462, opened 2022-08-18
Start the my rules page a bit

ryanlerch:

### DIFF
--- a/frontend/src/components/RuleList.vue
+++ b/frontend/src/components/RuleList.vue
@@ -1,0 +1,71 @@
+<script setup lang="ts">
+import {
+  CListGroup,
+  CListGroupItem,
+  CButton,
+  CModal,
+  CModalHeader,
+  CModalTitle,
+  CModalBody,
+} from "@coreui/bootstrap-vue";
+import type { QueryFunction } from "react-query/types/core";
+import { useQuery } from "vue-query";
+import { apiGet } from "../api";
+import { useUserStore } from "../stores/user";
+import NewRuleForm from "../components/NewRuleForm.vue";
+import RuleListItem from "../components/RuleListItem.vue";
+
+const userStore = useUserStore();
+const route = `/user/${userStore.username}/rules`;
+const { isLoading, isError, data, error } = useQuery(
+  route,
+  apiGet as QueryFunction<string[]>
+);
+</script>
+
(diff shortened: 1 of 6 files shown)


frontend/src/components/RuleListItem.vue
```diff
@@ -0,0 +1,22 @@
+<script setup lang="ts">
+import { CListGroupItem } from "@coreui/bootstrap-vue";
+const props = defineProps<{
+  rule: Object;
```
ryanlerch: I assume i have to make a type in types.ts for this

frontend/src/components/RuleList.vue
```diff
@@ -0,0 +1,71 @@
...
+  route,
+  apiGet as QueryFunction<string[]>
+);
+</script>
+
+<script lang="ts">
+export default {
+  data() {
+    return {
+      visibleNewRuleModal: false,
```
abompard: Instead of this `script` block, just define the variable in the `script setup` as a ref::
```
import { ref } from "vue";
const visibleNewRuleModal = ref(false);
```

frontend/src/components/RuleListItem.vue
abompard: yeah

frontend/src/components/RuleList.vue
```diff
@@ -0,0 +1,71 @@
...
+import { apiGet } from "../api";
+import { useUserStore } from "../stores/user";
+import NewRuleForm from "../components/NewRuleForm.vue";
+import RuleListItem from "../components/RuleListItem.vue";
+
+const userStore = useUserStore();
+const route = `/user/${userStore.username}/rules`;
+const { isLoading, isError, data, error } = useQuery(
+  route,
+  apiGet as QueryFunction<string[]>
```
abompard: The type after `QueryFunction` is what the API call will return. In your case it will be `Rule[]` when you have defined what a rule is, for now it can be `object[]` I suppose.

frontend/src/components/RuleListItem.vue
abompard: Also, the linter should tell you that in this case `Object` can be replaced by `unknown`.

abompard (changes requested):

PR #534, opened 2022-08-30
Sender component

abompard:
This is the start of the sender component, which will be instantiated once per destination, will listen to the internal queue, and send messages to users.


abompard:
OK, the sender component is now 100% tested. It does email (via an external SMTP server since we'll be running in OpenShift) and IRC.

PR #576, opened 2022-09-13
Add models

abompard:
These are some basic models for FMN's data. They will probably need optimization and indexes.


abompard:
Merged with #577

PR #579, opened 2022-09-14
Add logic so a user is created in the db on login to the frontend

ryanlerch:

### DIFF
--- a/fmn/api/main.py
+++ b/fmn/api/main.py
@@ -5,13 +5,20 @@ from fastapi import Depends, FastAPI, Request
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
 from pydantic import BaseModel
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession
 
 from ..core.config import Settings, get_settings
+from ..database import init_async_model
+from ..database.model import User
 from .auth import Identity, get_identity_optional
+from .database import req_db_async_session
 
 log = logging.getLogger(__name__)
 
 app = FastAPI()
+
+init_async_model()
 
 
 @app.on_event("startup")
@@ -23,6 +30,11 @@ def add_middlewares():
         allow_methods=["*"],
         allow_headers=["*"],
     )
+
+
+@app.on_event("startup")
+async def init_model():  # pragma: no cover todo
+    await init_async_model()
 
 
 class Rule(BaseModel):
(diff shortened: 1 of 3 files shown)


fmn/api/main.py
```diff
@@ -7,12 +7,17 @@
...
 from ..core.config import Settings, get_settings
+from ..database import init_sync_model, sync_session_maker
+from ..database.model import User
 from .auth import Identity, get_identity_optional
 
 log = logging.getLogger(__name__)
 
 app = FastAPI()
 
+init_sync_model()
```
abompard: Nope, FastAPI is async. Check out [their doc](https://fastapi.tiangolo.com/advanced/async-sql-databases/) about databases, they also recommend using events to connect to the database.

fmn/api/main.py
```diff
@@ -51,6 +56,17 @@ async def read_root(
     return result
 
 
+@app.get("/users/")
+def get_users():  # pragma: no cover
```
abompard: Look at the `read_root` function for arguments to require the user's credentials and identity.

fmn/database/model/user.py
```diff
@@ -8,3 +8,25 @@ class User(Base):
 
     id = Column(Integer, primary_key=True, nullable=False)
     name = Column(UnicodeText, nullable=False, unique=True)
+
+    @classmethod
+    def by_name(cls, session, name):  # pragma: no cover
+        return session.query(cls).filter_by(name=name).first()
```
abompard: Since 1.4, sqlalchemy recommends using the "2.0" style of making queries : https://docs.sqlalchemy.org/en/14/orm/queryguide.html
Also, it will have to be async.

fmn/database/model/user.py
```diff
@@ -8,3 +8,25 @@ class User(Base):
...
+    @classmethod
+    def create(cls, session, name):  # pragma: no cover
+        user = cls(name=name)
+        session.add(user)
+        session.commit()
+        return user
+
+    @classmethod
+    def all(cls, session):  # pragma: no cover
+        return session.query(cls).all()
```
abompard: Same here.

abompard (changes requested):

fmn/api/main.py
ryanlerch: read root wasnt working for me anymore on my tinystage instance -- i tired to see what was causing the issue but to no avail

ryanlerch:
Ok, it should now be interacting with the database asynchronously, and using the new style queries.

Still havent added auth to the new methods tho, as get root is still not working for me

PR #587, opened 2022-10-04
Make frontend CI happy

nphilipp:
- Remove unused variable
- Don't prohibit non-null assertion when linting
- Add missing type declarations
    
Signed-off-by: Nils Philippsen <nils@redhat.com>


abompard:
I did that in another PR, but thanks :-)

PR #599, opened 2022-10-11
Use listgroup for the generation rules list

ryanlerch:

### DIFF
--- a/frontend/src/components/rule-edit/generation-rule/GenerationRuleList.vue
+++ b/frontend/src/components/rule-edit/generation-rule/GenerationRuleList.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import type { GenerationRule } from "@/api/types";
-import { CButton, CCard, CCardBody } from "@coreui/bootstrap-vue";
+import { CButton, CListGroup, CListGroupItem } from "@coreui/bootstrap-vue";
 import { cilPlus } from "@coreui/icons";
 import { CIcon } from "@coreui/icons-vue";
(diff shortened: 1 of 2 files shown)


frontend/src/components/rule-edit/generation-rule/GenerationRuleList.vue
```diff
@@ -54,36 +54,35 @@ const editedRule = computed(() =>
...
+          @delete="() => handleRuleDeleted(index)"
+        />
+      </template>
+    </FormKit>
+    <CListGroupItem class="bg-light">
+      <h4 v-if="generation_rules.length === 0" class="text-center text-muted my-4">No Destinations Defined</h4>
+      <div class="text-center">
+      <CButton
+        @click.prevent="() => handleButtonClicked(-1)"
+        color="primary"
```
abompard: I had set the color to be secondary when there's already a rule because I thought it's a guideline to have only one primary button on the page (to make it clear what the next step is), and the Create Rule button is already primary. Do you think it's OK to have 2 primary buttons? I'll totally defer to your opinion on the matter, I just wanted to point this out.

abompard:
Also, please run the prettier formatter, it's what causes CI to fail. You can configure VSCode to run it on file save.

frontend/src/components/rule-edit/generation-rule/GenerationRuleList.vue
ryanlerch: +1 will add it back

Issue #609, opened 2022-10-12
Getting the list of rules fails


ryanlerch:
After this [commit](https://github.com/fedora-infra/fmn/commit/52c0735c7910731a9b0b61ff041fc98ac0ee9dd1#diff-ae920121865ed3ec31e759f19dccc0be0d4fa70a52c5884d129ed7a3f32fbf10), getting the list of rules fails with a 500 error:

```
Oct 12 04:39:27 fmn.tinystage.test poetry[8923]: INFO:     192.168.122.1:51792 - "GET /user/aaronhale/rules HTTP/1.1" 500 Internal Server Error
Oct 12 04:39:27 fmn.tinystage.test poetry[8923]: ERROR:    Exception in ASGI application
Oct 12 04:39:27 fmn.tinystage.test poetry[8923]: Traceback (most recent call last):
Oct 12 04:39:27 fmn.tinystage.test poetry[8923]:   File "/home/vagrant/.cache/pypoetry/virtualenvs/fmn-0n_c-6Cq-py3.10/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 404>
Oct 12 04:39:27 fmn.tinystage.test poetry[8923]:     result = await app(  # type: ignore[func-returns-value]
Oct 12 04:39:27 fmn.tinystage.test poetry[8923]:   File "/home/vagrant/.cache/pypoetry/virtualenvs/fmn-0n_c-6Cq-py3.10/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 78>
Oct 12 04:39:27 fmn.tinystage.test poetry[8923]:     return await self.app(scope, receive, send)
Oct 12 04:39:27 fmn.tinystage.test poetry[8923]:   File "/home/vagrant/.cache/pypoetry/virtualenvs/fmn-0n_c-6Cq-py3.10/lib/python3.10/site-packages/fastapi/applications.py", line 270, in __call>
Oct 12 04:39:27 fmn.tinystage.test poetry[8923]:     await super().__call__(scope, receive, send)
Oct 12 04:39:27 fmn.tinystage.test poetry[8923]:   File "/home/vagrant/.cache/pypoetry/virtualenvs/fmn-0n_c-6Cq-py3.10/lib/python3.10/site-packages/starlette/applications.py", line 124, in __ca>
Oct 12 04:39:27 fmn.tinystage.test poetry[8923]:     await self.middleware_stack(scope, receive, send)
[...]

Issue #606, opened 2022-10-11
Backend: Test all the things!!!


nphilipp:
# Story

As an FMN developer,
I want that all code is tested,
so I can be reasonably confident my changes don’t break anything.

# Acceptance Criteria

- [ ] Request handler functions etc. which do something beyond providing fake test data are tested and not marked `pragma: no cover`.
Non-exhaustive list:
  - [ ] `fmn.api.main`
    - [x] `init_model()` @nphilipp
    - [ ] `get_users()` @james02135 => in the context of #615 
    - [x] `get_user()` @james02135
    - [x] `get_user_info()`
    - [x] `get_user_groups()`
    - [x] `get_user_destinations()` @nphilipp
    - [x] `get_user_rules()` (#609) @nphilipp 
    - [x] `get_user_rule()`  @nphilipp
    - [x] `edit_user_rule()`  @nphilipp
    - [x] `delete_user_rule()` @nphilipp
    - [x] `create_user_rule()` @nphilipp
    - [x] `get_applications()` @nphilipp
    - [x] `get_owned_artifacts()` @nphilipp

PR #610, opened 2022-10-12
Fix previously broken handler functions and add tests

ryanlerch:
Got the list or rules to display again by returning a list.

Also then tried to add a new rule, and that failed too -- so started to try to fix that and failed. this is where i got to, and its probably all wrong.

Related: #609

### DIFF
--- a/fmn/api/main.py
+++ b/fmn/api/main.py
@@ -155,7 +155,7 @@ async def edit_user_rule(
     rule: api_models.Rule,
     identity: Identity = Depends(get_identity),
     db_session: AsyncSession = Depends(gen_db_session),
-):  # pragma: no cover todo
+):
     if username != identity.name:
         raise HTTPException(status_code=403, detail="Not allowed to edit someone else's rules")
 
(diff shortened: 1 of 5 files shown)


fmn/api/main.py
```diff
@@ -109,9 +109,8 @@ async def get_user_rules(
     if username != identity.name:
         raise HTTPException(status_code=403, detail="Not allowed to see someone else's rules")
 
-    return (
-        await db_session.execute(Rule.select_related().filter(Rule.user.has(name=username)))
-    ).scalars()
+    db_result = await db_session.execute(Rule.select_related().filter(Rule.user.has(name=username)))
+    return [Rule.from_orm(rule) for rule in db_result.scalars()]
```
nphilipp: We shouldn't have to manually convert to API model objects, `response_model=` above should take care of it. And plain `Rule` is the ORM class, not the API model class.

fmn/api/main.py
```diff
@@ -248,11 +247,9 @@ async def create_user_rule(
...
 
     # Refresh using the full query to get relationships
-    return (
-        await db_session.execute(
-            Rule.select_related().filter(Rule.id == id, Rule.user.has(name=username))
-        )
-    ).scalar_one()
+    db_result = await db_session.execute(Rule.select_related().filter(Rule.id == rule_db.id, Rule.user.has(name=username)))
+
+    return api_models.Rule.from_orm(db_result.scalar_one())
```
nphilipp: Same as above.

nphilipp (changes requested):
Also, let's do this in the context of #606.

nphilipp (approved):
LGTM (but I would say that 😉)

PR #612, opened 2022-10-13
Update the new rule page to disable rulename toggle

### LINKED ISSUE #601: Tweak editing rule name on New / Edit Rule Page
Currently on both the new rule and the rule edit form page, we present the rule name field as a label, with the pencil edit icon that when pressed shows the text field.

I'm thinking that on the new page, we should just show the field, as it pretty much will always be edited on that page. but keep the current pattern for the edit rule variant of the form

ryanlerch:
Resolves: https://github.com/fedora-infra/fmn/issues/601

### DIFF
--- a/frontend/src/components/NewRuleForm.vue
+++ b/frontend/src/components/NewRuleForm.vue
@@ -8,7 +8,6 @@ import { FormKit } from "@formkit/vue";
 import type { AxiosError } from "axios";
 import { computed, ref } from "vue";
 import { useRouter } from "vue-router";
-import EditableName from "./EditableName.vue";
 import GenerationRuleList from "./rule-edit/generation-rule/GenerationRuleList.vue";
 import TrackingRule from "./rule-edit/tracking-rule/TrackingRule.vue";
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/NewRuleForm.vue
```diff
@@ -66,9 +66,10 @@ const handleGenerationRulesChanged = (rules: GenerationRule[]) => {
         <h4>
           <EditableName
             name="name"
-            value="New Rule 1"
-            button-class=""
-            input-class="form-control-lg"
+            value=""
+            button-class="fs-3"
```
abompard: The `fs-3` class here makes the icon a bit lower than the text. Removing the class makes it look properly aligned. How about removing it?

frontend/src/components/EditableName.vue
```diff
@@ -30,26 +31,38 @@ onUpdated(() => {
 </script>
 
 <template>
-  <span :class="{ 'd-none': !editing }">
+  <template v-if="props.noToggle === true">
```
abompard: I don't think it's worthwhile to regroup both toggle and noToggle modes in the same component, since they are displaying very different things and no code is actually shared between the two modes, is it?
I would just put the `FormKit` element in the `NewRuleForm`, since it's the only place where it's called with `noToggle`.

ryanlerch: you are right -- i was def ovethinking this one a bit :)

have updated it now to just use a text input vias formkit

PR #615, opened 2022-10-13
Query FASjson for a list of users to track

### LINKED ISSUE #614: Query FASJSON to add additional users other than yourself to the artifacts owned tracking rule
Currently, the "artifacts_owned" parameter for users uses the "/users/" endpoint, which currently queries the fmn database for a list of users.

However, i'm pretty sure we want to query the fasjson list of users here, as someone may want to track the artifacts owned by a user that has never used FMN.

ryanlerch:
Query FASjson for a list of users to track

Resolves: #614

### DIFF
--- a/fmn/api/main.py
+++ b/fmn/api/main.py
@@ -5,7 +5,6 @@ from fastapi import Depends, FastAPI, HTTPException, Query, Request, status
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
-from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
 from starlette.middleware.errors import ServerErrorMiddleware
 from starlette.requests import Request as StarletteRequest
@@ -72,12 +71,11 @@ async def read_root(
     return result
 
 
-@app.get("/users/", response_model=list[api_models.User])
-async def get_users(
-    db_session: AsyncSession = Depends(gen_db_session),
+@app.get("/users/", response_model=list[str])
+def get_users(
+    search: str, fasjson_client: FasjsonClient = Depends(get_fasjson_client)
 ):  # pragma: no cover todo
-    result = await db_session.execute(select(User))
-    return result.scalars().all()
+    return [u["username"] for u in fasjson_client.search(username=search).result]
 
 
 @app.get("/user/{username}/", response_model=api_models.User)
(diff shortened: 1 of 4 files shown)


fmn/api/main.py
```diff
@@ -72,20 +72,25 @@ async def read_root(
...
-    return {"users": result.scalars().all()}
+    return result.scalars().all()
 
 
-@app.get("/user/{username}/")
-async def get_user(
-    username, db_session: AsyncSession = Depends(gen_db_session)
+@app.get("/fasjsonusers/")
+def get_fasjsonusers(
+    fasjson_client: FasjsonClient = Depends(get_fasjson_client),
```
abompard: This should have a `response_model` too, no?

frontend/src/components/rule-edit/tracking-rule/ArtifactsOwnedParams.vue
```diff
@@ -3,43 +3,50 @@ import { apiGet } from "@/api";
...
-    :min-chars="3"
-    :delay="0"
-  />
-  <ArtifactsOwnedSummary :users="value" />
+  <p v-if="isLoading">
+    <input
+      type="text"
+      class="form-control"
+      disabled
+      value="Loading groups..."
```
abompard: It's actually loading users I think.

frontend/src/components/rule-edit/tracking-rule/ArtifactsOwnedParams.vue
```diff
@@ -3,43 +3,50 @@ import { apiGet } from "@/api";
...
+    />
+  </p>
+  <span v-else-if="isError">Could not load groups: {{ error }}</span>
```
abompard: Same here

fmn/api/main.py
```diff
@@ -72,20 +72,25 @@ async def read_root(
...
 ):  # pragma: no cover todo
+    return [u["username"] for u in fasjson_client.list_all_entities("users")]
```
abompard: This is going to take way too long. I think it should forward a search for the username to fasjson.

frontend/src/components/rule-edit/tracking-rule/ArtifactsOwnedParams.vue
```diff
@@ -3,43 +3,50 @@ import { apiGet } from "@/api";
...
-    return value.value;
-  }
-  const results = await apiGetUsers({
-    queryKey: [route, { search: query }],
-    meta: undefined,
-  });
-  return results ? results.users.map((u) => u.name) : [];
-};
+const url = "/fasjsonusers/";
+const { isLoading, isError, data, error } = useQuery(url, apiGetUsers);
```
abompard: As commented above, I don't think that's the right approach, it should search for the partial username instead of loading them all from fasjson, as there will be too many of them.

abompard (dismissed):

nphilipp:
Rebasing on top of #616 so that you don't have to jump through hoops to test the new, proxying-FASJSON `get_users()`.

nphilipp (approved):
Looks good to me!

PR #619, opened 2022-10-21
Production fixes

Zlopez:
These are the fixes for the current production deployment.


Zlopez:
I will ignore the diff cover, because there aren't even tests files for most source files changed in this PR.

PR #627, opened 2022-10-27
Make the rules and associated objects accessible from the API

abompard:
This is necessary to provide previsualisation of a rule in the UI.

The tracking rules and filters are still synchronous and will have to be run in a thread in the API.


fmn/testing/message.py
```diff
@@ -1,9 +1,6 @@
 import pytest
 from fedora_messaging import message
```
nphilipp: What's the reason to move this from `tests/` into the `fmn` package (where it will be installed)? We can have importable modules inside `tests/`…

tests/conftest.py
```diff
@@ -20,6 +20,8 @@
 )
 from fmn.database.model import Destination, Filter, GenerationRule, Rule, TrackingRule, User
 
+pytest_plugins = "fmn.testing"
+
```
nphilipp: This applies the above `register_message_class(autouse=True, ...)` fixture in all tests, not just those beneath `tests/consumer`, is that intentional?

fmn/testing/message.py
abompard: Ah, how do you do that? I tried to put the message class in a module at the top of the `tests` tree, but `from ...utils import Message` failed.

tests/conftest.py
abompard: It's for the consumer tests and the rules tests now. I guess I could add it to `conftests.py` modules inside each, but the `register_message_class` fixture is pretty harmless. We may want to use this message class in the API too at some point when implementing the filtering preview feature anyway.

fmn/testing/message.py
abompard: To be precise I get: `ImportError: attempted relative import beyond top-level package`

abompard: OK, in a followup commit I've changed how mock messages are created, and it gets rid of the `fmn.testing` module.

I'll squash when the PR is approved.

nphilipp (approved):
LGTM!

PR #628, opened 2022-10-27
Get Artifacts followed working with dummy endpoint

ryanlerch:


fmn/api/handlers/misc.py
```diff
@@ -40,3 +40,16 @@ def _add_artifact(a):
...
+def get_artifacts(type: str, name: str):  # pragma: no cover todo
+    # TODO: Get artifacts
+
+    artifacts = []
+
+    artifacts.append(f"foobar-{name}")
+    artifacts.append(f"foo-{name}")
+    artifacts.append(f"bar-{name}")
+
+    return artifacts
```
nphilipp, line 55: This looks like we could consolidate this all in an `/artifacts` endpoint with a set of query parameters to restrict the result set (and a uniform result format). What do you think?

PR #640, opened 2022-11-07
query distgit for artifacts

### LINKED ISSUE #603: Get the list of owned artifacts by a user and/or a group

ryanlerch:
Filed early for feedback on this approach.

still need to tweak how we search for artifacts type (rpm/module/flatpak etc), and need to tweak the frontend too.

it queries the new src.tinystage.test so youll need to provision that in tinystage.

### DIFF
--- a/devel/ansible/roles/backend/templates/fmn.cfg.j2
+++ b/devel/ansible/roles/backend/templates/fmn.cfg.j2
@@ -6,3 +6,4 @@ OIDC_TOKEN_INFO_ENDPOINT="/TokenInfo"
 OIDC_CLIENT_ID="{{ oidc_client_id }}"
 OIDC_CLIENT_SECRET="{{ oidc_client_secret }}"
 SERVICES__FASJSON_URL=http://fasjson.tinystage.test/fasjson
+SERVICES__DISTGIT_URL=http://src.tinystage.test
--- a/fmn/api/distgit.py
+++ b/fmn/api/distgit.py
@@ -1,0 +1,31 @@
+import logging
+
+from fastapi import Depends
+from httpx import AsyncClient
+
+from ..core.config import Settings, get_settings
+
+log = logging.getLogger(__name__)
+
+
+class DistGitClient:
+    def __init__(self, settings):
+        self.client = AsyncClient(base_url=settings.services.distgit_url, timeout=None)
+
+    async def get_artifacts(self, pattern):
+        response = await self.client.get(
+            "/api/0/projects",
+            params={
+                "pattern": f"*{pattern}*",
+                "short": "true",
+                "fork": "false",
+            },
+        )
+        response.raise_for_status()
+        result = response.json()
+
+        return [project["name"] for project in result["projects"]]
+
+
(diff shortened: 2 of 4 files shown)


devel/ansible/roles/backend/templates/fmn.cfg.j2
```diff
@@ -6,3 +6,4 @@ OIDC_TOKEN_INFO_ENDPOINT="/TokenInfo"
 OIDC_CLIENT_ID="{{ oidc_client_id }}"
 OIDC_CLIENT_SECRET="{{ oidc_client_secret }}"
 SERVICES__FASJSON_URL=http://fasjson.tinystage.test/fasjson
+SERVICES__DISTGIT_URL=http://src.tinystage.test/
```
abompard: You should probably leave out the last `/`, otherwise you'll have two slashes when you build the full URL in the client.

fmn/api/distgit.py
```diff
@@ -0,0 +1,25 @@
+import logging
+
+import requests
```
abompard: This is synchronous, it would be better to use httpx's `AsyncClient`.  See `fmn/api/auth.py` for an example.

fmn/api/distgit.py
```diff
@@ -0,0 +1,25 @@
...
+class DistGitClient:
+    def __init__(self, settings):
+        self.url = settings.services.distgit_url
+        self.session = requests.Session()
+
+    def get_artifacts(self, pattern):
+        url = f"{self.url}/api/0/projects"
+        projects = self.session.get(
+            url=url, params={"pattern": f"*{pattern}*", "short": "true"}
+        ).json()
```
abompard: Please use `response.raise_for_status()` to get an exception when we can't reach distgit. You'll call `.json()` after that. See `fmn/api/auth.py` for an example.

abompard (changes requested):

ryanlerch:
@abompard thanks! have implementied your reccomendations!

ryanlerch:
Ok, tried and failed to add some tests -- could not get the mocking for the async calls working properly :(

ryanlerch:
Got the tests working, so this one is ready for final review!

fmn/api/distgit.py
```diff
@@ -0,0 +1,30 @@
...
+    async def get_artifacts(self, pattern):
+        response = await self.client.get(
+            "/api/0/projects",
+            params={
+                "pattern": f"*{pattern}*",
+                "short": "true",
+            },
+        )
+        response.raise_for_status()
+        result = response.json()
```
abompard, line 25: In the results I get all the forks of the search projects. Is there a way to query distgit to exclude the forks? (I tried searching for `0`).

abompard (changes requested):

fmn/api/distgit.py
ryanlerch: whoops! i forgot to add this in (my test data doesnt have any forks, so missed this)

ryanlerch: fixed!

PR #645, opened 2022-11-08
Isolate unit tests from network

### LINKED ISSUE #635: Isolate unit tests
# Story

As a developer of FMN,
I want that tests don’t attempt to contact external systems,
so results are consistent and deterministic.

# Acceptance Criteria

- [x] Network connections using the `requests` package are intercepted across the board using an appropriate fixture.
- [x] Fixtures are implemented or extended so often-used endpoints return something sensible.

# Background

Outside of the web API, we use `fasjson_client` which expects certain OpenAPI-related endpoints to work (i.e. one that specs out the FASJSON API (`http://fasjson.…/specs/v1.json`) which seems to reference some more on `https://json-schema.org`)

nphilipp:
We use the `responses` package to mock out HTTP traffic to external systems which use the `requests` package. However, FASJSON calls were only configured to be mocked out for some tests, others used passthrus which let them communicate with the Fedora FASJSON service and json-schema.org. This fails every now and then when there's some hickup on the network or one of the involved systems.

This consolidates mocking out requests to FASJSON and related services in the `mocked_fasjson` fixture which is made available globally.

Fixes: #635

Signed-off-by: Nils Philippsen <nils@redhat.com>


nphilipp:
No review is silent assent or something like that. 😉

PR #648, opened 2022-11-10
Properly display owned artifacts by querying distgit

ryanlerch:

### DIFF
--- a/fmn/api/distgit.py
+++ b/fmn/api/distgit.py
@@ -1,4 +1,5 @@
 import logging
+from urllib.parse import urlencode
 
 from fastapi import Depends
 from httpx import AsyncClient
@@ -24,6 +25,35 @@ class DistGitClient:
         response.raise_for_status()
         result = response.json()
         return result["projects"]
+
+    async def get_owned(self, names, user_or_group):
+        def _append_artifacts(projects):
+            for p in projects:
+                artifacts.append({"type": p["namespace"], "name": p["name"]})
+
+        artifacts = []
+        endpoint = "/api/0/projects"
+        for name in names:
+            params = {"short": "true", "fork": "false"}
+
+            if user_or_group == "user":
+                params["owner"] = name
+            elif user_or_group == "group":
+                params["username"] = f"@{name}"
+            else:
+                raise ValueError("Argument user_or_group must be either user or group")
+
+            params["page"] = "1"
+            url = f"{endpoint}?{urlencode(params)}"
+
+            while url:
+                response = await self.client.get(url)
+                response.raise_for_status()
+                data = response.json()
+                _append_artifacts(data["projects"])
+                url = data.get("pagination", {}).get("next")
+
+        return artifacts
 
 
 def get_distgit_client(settings: Settings = Depends(get_settings)):
--- a/tests/api/test_handlers.py
+++ b/tests/api/test_handlers.py
@@ -312,35 +312,66 @@ class TestMisc(BaseTestAPIV1Handler):
             "planet",
         ]
 
-    @pytest.mark.parametrize(
-        "testcase", ("query-users", "query-users-duplicate", "query-groups", "query-nothing")
-    )
-    def test_get_owned_artifacts(self, testcase, client):
-        # get_owned_artifacts() looks straw-manny, so just check structure
+    @pytest.mark.parametrize("ownertype", ("user", "group"))
+    async def test_get_projects(self, client, mocker, async_respx_mocker, ownertype):
+        settings = get_settings()
+        settings.services.distgit_url = "http://distgit.test"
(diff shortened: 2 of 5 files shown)


abompard:
Could you rebase and fix the conflict please?

ryanlerch:
> Could you rebase and fix the conflict please?

done.

not sure what the failing test is from, but its happening in the main branch too

fmn/api/distgit.py
```diff
@@ -25,6 +26,37 @@ async def get_projects(self, pattern):
...
+            elif user_or_group == "group":
+                params["username"] = f"@{name}"
+            else:
+                raise ValueError("Argument user_or_group must be either user or group")
+
+            firstpage = await self.client.get(endpoint, params=params)
+            firstpage.raise_for_status()
+            _appendArtifacts(firstpage.json()["projects"])
+
+            if firstpage.json()["pagination"]["pages"] > 1:
```
abompard: You could simplify that with a `while` loop. Check out `fmn.rules.services.distgit:DistGitService._all_values` for an example.

abompard (changes requested):

fmn/api/distgit.py
ryanlerch: okay... so i implemented this -- and it worked -- however it started to make the tests a lot more difficult
here is the branch with the change, and my attempts at redoing the tests

https://github.com/ryanlerch/fmn/tree/artifacts2-while-fullurl

basically, i previously used the same response twice with return_value, however, getting async coroutine StopIteration issues when trying to use side_effect (returning a list) to do multiple different return values.

is there something like responses for httpx async here to make this easier?

ryanlerch: I think i have ti sorted -- PR changes incoming soon

ryanlerch:
Ok! using the "next" url returned by pagure now in a while loop.

Got the tests working now too with pytest-httpx to mock the responses which works a lot better than what i was doing before.

Kept the initial commit in there, but intend to squash them into 1 before merging

fmn/api/distgit.py
```diff
@@ -25,6 +25,36 @@ async def get_projects(self, pattern):
...
+                params["owner"] = name
+            elif user_or_group == "group":
+                params["username"] = f"@{name}"
+            else:
+                raise ValueError("Argument user_or_group must be either user or group")
+
+            params["page"] = "1"
+            response = await self.client.get(endpoint, params=params)
+            response.raise_for_status()
+            print(response.json())
```
abompard: Leftover here :-)

fmn/api/distgit.py
```diff
@@ -25,6 +25,36 @@ async def get_projects(self, pattern):
...
+
+            _appendArtifacts(response.json()["projects"])
+            while response.json().get("pagination", {}).get("next"):
+                response = await self.client.get(response.json()["pagination"]["next"])
```
abompard: Yeah that works. Another way of doing it would have been to initialize `url` with the initial URL and refresh it with the next URL in the response, such as:
```python
url = f"{endpoint}?{urlencode(params)}"
while url:
    response = await self.client.get(url)
    response.raise_for_status()
    data = response.json()
    _appendArtifacts(data["projects"])
    url = data.get("pagination", {}).get("next")
```
This would allow you to avoid repeating yourself when writing the `client.get` calls, the `raise_for_status`, `_appendArtifacts` and such.

fmn/api/distgit.py
```diff
@@ -25,6 +25,36 @@ async def get_projects(self, pattern):
         result = response.json()
         return result["projects"]
 
+    async def get_owned(self, names, user_or_group):
+        def _appendArtifacts(projects):
```
abompard: That's javascript naming scheme, please use `_append_artifacts`.

tests/api/test_handlers.py
```diff
@@ -285,35 +286,51 @@ def test_get_applications(self, client):
...
+        )
 
-        response = client.get(f"{self.path}/artifacts/owned", params=params)
+        if ownertype == "user":
+            response = client.get(f"{self.path}/artifacts/owned", params={"users": ["dudemcpants"]})
+        elif ownertype == "group":
+            response = client.get(f"{self.path}/artifacts/owned", params={"groups": ["dudegroup"]})
 
-        assert response.status_code == status.HTTP_200_OK
+        print(httpx_mock.get_requests())
```
abompard: Not sure you need that either.

abompard (changes requested):

ryanlerch:
Thanks for the review!

Have implemented your changes, and squashed the commits now too.

ryanlerch:
Okay! Updated the tests to use respx (which was already being used) rather than pytest-httpx (which i introduced in a previous iteration of thius PR before knowing about respx)

abompard (approved):

PR #652, opened 2022-11-14
Fix some deps issues

nphilipp:


nphilipp:
LGTM but I would say that :smirk:.

PR #657, opened 2022-11-16
Add s2i integration

abompard:
This adds the files necessary for Source 2 Image integration, that OpenShift will use to build the image to run in pods.

For testing, build the image with:
```
s2i build . quay.io/fedora/python-310 fmn-test --copy --rm
```

There will be 3 possible ways to run the image:

- with `APP_SCRIPT=.s2i/run-api.sh` to run the API
- with `APP_SCRIPT=.s2i/run-consumer.sh` to run the consumer
- with `APP_SCRIPT=.s2i/run-sender.sh` to run the sender

### DIFF
--- a/.s2i/bin/assemble
+++ b/.s2i/bin/assemble
@@ -1,0 +1,19 @@
+#!/bin/bash -e
+
+# The assemble script builds the application artifacts from a source and
+# places them into appropriate directories inside the image.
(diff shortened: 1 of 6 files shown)


ryanlerch (approved):
LGTM

Successfully built the images locally.

PR #668, opened 2022-11-25
Redis role added in devel/ansible, redis playbook and config file added

### LINKED ISSUE #667: Redis role added to devel/ansible

james02135:
Signed-off-by: james02135 <james02135@hotmail.com>

### DIFF
--- a/devel/ansible/playbook.yml
+++ b/devel/ansible/playbook.yml
@@ -18,3 +18,4 @@
     - frontend
     - consumer
     - sender
+    - redis
--- a/devel/ansible/roles/redis/tasks/main.yml
+++ b/devel/ansible/roles/redis/tasks/main.yml
@@ -1,0 +1,13 @@
+---
+- name: Install RPM packages
+  dnf:
+      name:
+        - redis
+      state: present
+      
+- name: Start the service using systemd
+  systemd:
+    state: started
+    name: redis
+    daemon_reload: yes
+    enabled: yes


abompard (dismissed):
I'm getting this error:
```
$ vagrant provision
==> fmn: Running provisioner: ansible...
    fmn: Running ansible-playbook...
[...]
ERROR! conflicting action statements: hosts, tasks

The error appears to be in '[...]/devel/ansible/roles/redis/tasks/main.yml': line 1, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:


- hosts: redis
  ^ here
Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
```

The syntax of the `tasks/main.yml` file is not the same as a playbook, it's basically only tasks. Handlers definition go to `handlers/main.yml`. Check out [this doc](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html) for reference.

nphilipp (changes requested):
The redis role should install the redis RPM package, without it, it fails here when I `vagrant up --provision`:

```
TASK [redis : Start the service using systemd] *********************************
fatal: [fmn]: FAILED! => changed=false 
  msg: 'Could not find the requested service redis: host'
```

Also, the redis configuration file in the role isn’t installed – it might not have to be as the RPM package comes with one (could work for us out of the box or not, let’s see).

And the usual "use imperative mood for commit logs" 😉, cf. https://cbea.ms/git-commit/#seven-rules

nphilipp (approved):

PR #686, opened 2022-12-07
Update vagrant setup to use the fmn vhost recently added to tinystage

ryanlerch:


devel/ansible/roles/consumer/files/config.toml
```diff
@@ -1,4 +1,4 @@
-amqp_url = "amqps://fedora:@rabbitmq.fedoraproject.org/%2Fpublic_pubsub"
+amqp_url = "amqp://fedoramessages:fedoramessages@tinystage.tinystage.test"
 callback = "fmn.consumer:Consumer"
 
 [tls]
```
abompard, line 4: We can drop this section as well, no?

PR #702, opened 2022-12-14
Have common API proxy classes usable from async and sync code

### LINKED ISSUE #690: Consolidate backend service wrappers: async & sync
# Story

As an FMN developer,
I want to be able to implement backend wrappers only once,
because I **Don’t Want to Repeat Myself™**.

# Acceptance Criteria

- [x] A mechanism exists which lets me call arbitrary async functions/methods from synchronous code
- [x] Backend wrappers employ this mechanism to make their functionality available for “both sides of the \[sync/async\] fence”.
  - [x] fasjson
  - [x] distgit/Pagure
  - [x] datagrepper
- [x] These wrappers are used across the codebase

nphilipp:
This allows us to consolidate caching in one place. 

Fixes: #690


tests/backends/test_base.py
```diff
@@ -0,0 +1,146 @@
...
+
+        client.client.get.assert_awaited_once_with("url", foo="bar")
+        response.raise_for_status.assert_called_once_with()
+        assert result is sentinel
+
+    async def test_get_payload(self, client):
+        with mock.patch.object(client, "get") as client_get:
+            client_get.return_value = {"result": "boo"}
+
+            result = await client.get_payload("boo")
```
abompard: Ah, this boo is the URL, not the payload value. For a moment I was scared that it might have been ;-)

fmn/backends/datagrepper.py
```diff
@@ -0,0 +1,38 @@
...
+
+class DatagrepperAsyncProxy(APIClient):
+    """Proxy for the Datagrepper API endpoints used in FMN"""
+
+    API_VERSION = "v2"
+
+    payload_field = "raw_messages"
+
+    def __init__(self, base_url: str):
+        self.client = AsyncClient(base_url=f"{base_url.rstrip('/')}/{self.API_VERSION}", timeout=0)
```
abompard: Actually the timeout should be `None` to disable it. When it's zero it timeouts instantly :-)

fmn/backends/pagure.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class PagureAsyncProxy(APIClient):
+    """Proxy for the FASJSON API endpoints used in FMN"""
+
+    API_VERSION = "0"
+
+    def __init__(self, base_url: str):
+        self.client = AsyncClient(
+            base_url=f"{base_url.rstrip('/')}/api/{self.API_VERSION}", timeout=0
```
abompard: Same thing here, the timeout should be `None` to disable it. https://www.python-httpx.org/advanced/#timeout-configuration

fmn/api/handlers/utils.py
```diff
@@ -39,25 +39,10 @@ def gen_requester() -> Iterator[Requester]:
...
-        )
-        response.raise_for_status()
-        data = response.json()
-        for msg_dict in data["raw_messages"]:
-            yield get_message(msg_dict)
-        total_pages = data["pages"]
-        if page >= total_pages:
-            break
-        page += 1
+    proxy = DatagrepperSyncProxy(get_settings().services.datagrepper_url)
```
abompard, line 42: When testing this out I get:
`RuntimeError: <asyncio.locks.Event object at 0x7fbd1da4c9a0 [unset]> is bound to a different event loop`
Maybe FastAPI does not like having an event loop running in a thread called by another event loop. But in theory we should be able to use the async proxy here, no?

abompard (changes requested):
Great work, thanks! I found a couple small bugs while testing it out, but otherwise it seems to work :-)

fmn/api/handlers/utils.py
nphilipp: Yeah, meanwhile the glue code (`get_last_messages()`) is so small we could simply use the sync and async versions of the API proxy.

fmn/backends/datagrepper.py
nphilipp: Ahh, good catch.

tests/backends/test_base.py
nphilipp: I probably should make it look like a URL so that people aren’t confused 😀.

fmn/api/handlers/utils.py
nphilipp: Hmm, this is used in `preview_rule()` which indirectly uses a `Requester` (which is sync). I'll give it a shot with leaving this code as is but let the wrapper attempt to get the existing asyncio event loop and only create its own one on failure.

nphilipp: > Hmm, this is used in `preview_rule()` which indirectly uses a `Requester` (which is sync). I'll give it a shot with leaving this code as is but let the wrapper attempt to get the existing asyncio event loop and only create its own one on failure.

The latest push contains this change.

abompard (approved):
Yep, looks good and seems to work, thanks a lot! :-)

PR #707, opened 2022-12-15
Add a switch to the edit rule page to disable rules

ryanlerch:

### DIFF
--- a/frontend/src/components/RuleEditForm.vue
+++ b/frontend/src/components/RuleEditForm.vue
@@ -102,9 +102,15 @@ const formReady = computed(() => generationRulesCount.value > 0);
     :actions="false"
     :value="props.rule"
   >
-    <CRow class="mb-2">
-      <CCol xs="auto" class="me-auto">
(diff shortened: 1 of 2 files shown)


frontend/src/components/RuleEditForm.vue
```diff
@@ -102,9 +104,16 @@ const formReady = computed(() => generationRulesCount.value > 0);
...
   >
-    <CRow class="mb-2">
-      <CCol xs="auto" class="me-auto">
-        <h4>
+    <CRow class="mb-2 d-flex">
+      <CCol xs="auto" class="flex-grow-1 d-flex align-items-center h3">
+        <FormKit
+          type="checkbox"
+          name="disabled"
+          :v-model="disabledValue"
```
abompard: You don't actually need this line. FormKit will use the field name to match the incoming and outgoing data. (you therefore also don't need to define `disabledValue` at all)

abompard (changes requested):

PR #709, opened 2022-12-21
UI improvements

ryanlerch:
A few UI improvements / bugfixes -- each commit will be a separate thing

### DIFF
--- a/frontend/src/components/RuleListItem.vue
+++ b/frontend/src/components/RuleListItem.vue
@@ -43,11 +43,9 @@ const tracking_rule = TRACKING_RULES.find(
         >
         <template
           v-if="
-            [
-              'artifacts-owned',
-              'artifacts-group-owned',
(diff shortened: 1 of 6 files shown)


frontend/src/components/RuleListItem.vue
```diff
@@ -21,15 +21,52 @@ const tracking_rule = TRACKING_RULES.find(
...
         >
       </div>
       <div>
-        {{ tracking_rule.label }}
+        {{ tracking_rule.prefixlabel
+        }}<template
+          v-if="
+            rule.tracking_rule.params && rule.tracking_rule.params.length > 1
+          "
+          >s:</template
```
abompard, line 37: This technique does not seem very i18n-friendly, but I suppose we can start with that.

abompard (approved):
Let's start with that! :-)

PR #717, opened 2023-01-09
Refactor querying Distgit/Pagure owners (and caching)

### LINKED ISSUE #689: Cache requests to distgit (Pagure)
# Story

As a user of FMN,
I want that data retrieved from distigt (Pagure) is appropriately cached,
so that this backend system isn’t hammered too hard.

# Acceptance Criteria

- [x] Allow configuring distgit caching lifetime
- [x] Data retrieved from distgit is cached

nphilipp:
Apart from reworking how FMN accesses Pagure/Distgit, plus quite some minor improvements and fixes. Sorry for the huge drop! 😂


fmn/api/handlers/misc.py
```diff
@@ -80,8 +86,7 @@ async def get_artifacts(
         },
     ]
     namespaces = [at.value for at in ArtifactType]
-    projects = await distgit_client.get_projects(pattern=name)
-    for project in projects:
+    for project in await distgit_client.get_projects(pattern=f"*{name}*"):
```
abompard: Are you sure we need to do a substring search by padding asterisks? It feels like we may end up with more projects than we expect.

fmn/rules/services/distgit.py
```diff
@@ -19,16 +19,17 @@ def __init__(self, url):
...
-        if user_or_group == "user":
-            return result["access_users"]["owner"]
-        elif user_or_group == "group":
-            return result["access_groups"]["admin"] + result["access_groups"]["commit"]
-        else:
-            raise ValueError("Argument user_or_group must be either user or group, duh.")
+    def get_user_owners(self, artifact_type: str, artifact_name: str) -> list[str]:
+        return self.proxy.get_project_users(
+            f"{artifact_type}/{artifact_name}",
+            PagureRole.OWNER | PagureRole.ADMIN | PagureRole.COMMIT,
```
abompard: Huh, I did not, know that OR'ed IntEnums could be iterated on to get the components!

In fact, after some research, it seems to be a new feature of Python 3.11. Can you confirm? I think we want to target 3.10, no?

fmn/api/handlers/misc.py
nphilipp: This was done in the (`fmn.rules`) distgit client previously, I moved that out of the implementation in `fmn.backends`. I'm fine with removing it altogether or move this detail into the frontend (people will probably want to search by substring, I know I want to).

fmn/rules/services/distgit.py
nphilipp: You can iterate over the type and get the constituent enum values, this is what's done in `PagureAsyncProxy.get_project_users()` and it seems to work in Python 3.10, too (same output in Python 3.11):

```
nils@makake:~> python3.10
Python 3.10.9 (main, Dec 20 2022, 00:00:00) [GCC 12.2.1 20221121 (Red Hat 12.2.1-4)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from enum import IntEnum, auto
>>> class Foo(IntEnum):
...     bar = auto()
...     baz = auto()
...     gna = auto()
... 
>>> Foo.bar | Foo.baz 
3
>>> for x in Foo:
...     print(x)
... 
Foo.bar
Foo.baz
Foo.gna
>>>
```

.github/workflows/main.yaml
```diff
@@ -1,6 +1,27 @@
...
+      - uses: actions/checkout@v3
+
+      - name: Install absolufy
+        run: |
+             python3 -m pip install absolufy-imports
+
+      - name: Enforce relative imports in package
+        run: >
+             find fmn -name \*.py -a \! -path fmn/database/migrations/env.py -print0
+             | xargs -0 absolufy-imports --never
```
abompard, line 21: Why is the Alembic script excluded and then handled differently? I haven't (easily) found what the `--never` switch does.

fmn/backends/pagure.py
```diff
@@ -30,5 +55,187 @@ def determine_next_page_params(self, url: str, params: dict, result: dict) -> Ne
...
+                if role & acl and role.bit_count() == 1
+            ]
+
+        seen_fullnames = set()
+        sorted_projects = []
+        for params in params_seq:
+            async for project in self.get_paginated(
+                f"/group/{name}", params=params, payload_field="projects"
+            ):
+                if (fullname := project["fullname"]) in seen_fullnames:
```
abompard, line 133: Ha, I think that's the first time I see `:=` used! :-)

fmn/backends/pagure.py
```diff
@@ -30,5 +55,187 @@ def determine_next_page_params(self, url: str, params: dict, result: dict) -> Ne
...
+                    get_pattern_for_cached_calls(
+                        self.get_projects, self=self, username=None, owner=None
+                    ),
+                    get_pattern_for_cached_calls(self.get_projects, self=self, username=user),
+                    get_pattern_for_cached_calls(self.get_projects, self=self, owner=user),
+                )
+                async for key, _ in cache.get_match(pattern)
+            ]
+        else:  # usergroup == "group"
+            # Messages about changes to project users
```
abompard: to project *groups*, no?

abompard (commented):
That's a lot of work! Thanks!
Only minor comments. There are some parts with invalidation and the computation of key templates that I'm not sure I understood entirely, so I'll read it over again, but I'm happy testing it on staging anyway.

fmn/backends/pagure.py
nphilipp: Coming from C and the like, I was missing it so much in Python! 🙂

.github/workflows/main.yaml
nphilipp: - Alembic imports the `env.py` file and migration scripts in a way that they don’t know they are inside the `fmn` package, any relative imports outside of the `fmn/database/migrations` directory would fail. In `env.py` there’s a comment about it, I’ll add one here, too.
- The default mode of `absolufy-imports` is to enfore imports are absolute instead of relative, `--never` does the opposite.

fmn/backends/pagure.py
nphilipp: Oops, of course!

abompard (approved):
Alright, let's merge this and see what happens :-D

PR #723, opened 2023-01-18
Switch the cache to cashews and make the consumer async

abompard:
Huge change. Use the cashews library to have an async cache.
Also switch the consumer to asyncio because it's too complex to bridge sync code and asyncio when there are multiple async entry points.


fmn/api/handlers/utils.py
```diff
@@ -33,15 +33,15 @@ def gen_requester() -> Iterator[Requester]:
     :return: A :class:`fmn.rules.requester.Requester` object for the
         current request
     """
-    requester = Requester(get_settings().dict()["services"])
+    requester = Requester(get_settings().services)
```
nphilipp, line 36: 👍

fmn/api/handlers/utils.py
```diff
@@ -33,15 +33,15 @@ def gen_requester() -> Iterator[Requester]:
...
 
 
 # TODO: absolutely cache this
-def get_last_messages(hours):
-    proxy = DatagrepperSyncProxy(get_settings().services.datagrepper_url)
+async def get_last_messages(hours):
+    proxy = DatagrepperAsyncProxy(get_settings().services.datagrepper_url)
 
-    for msg_dict in proxy.search(delta=int(hours) * 60 * 60, rows_per_page=100):
+    async for msg_dict in proxy.search(delta=int(hours) * 60 * 60, rows_per_page=100):
```
nphilipp, line 44: I was puzzled by this one (mainly because no type hints – hint 😉!), couldn't `get_last_messages()` accept `hours: int | float` and cast it to seconds as `int(hours * 60 * 60)`?

Could even accept `delta: timedelta | None` alternatively, but that'd be probably overengineered for our purposes 😀.

fmn/cache/tracked.py
```diff
@@ -0,0 +1,73 @@
...
+from typing import TYPE_CHECKING
+
+from cashews import cache
+from cashews.formatter import get_templates_for_func
+
+from fmn.database.model import Rule
+
+if TYPE_CHECKING:
+    from fedora_messaging.message import Message
+    from sqlalchemy.orm import Session
```
nphilipp: This should be …:
```
    from sqlalchemy.ext.asyncio import AsyncSession
```

fmn/cache/tracked.py
```diff
@@ -0,0 +1,73 @@
...
+        tracked = Tracked()
+        db_result = await db.execute(Rule.select_related().filter_by(disabled=False))
+        for rule in db_result.scalars():
+            await rule.tracking_rule.prime_cache(tracked, requester)
+        log.debug("Built the tracked cache")
+        return tracked
+
+    @cache.early(key="tracked", prefix="v1", ttl="1d", early_ttl="23h")
+    # @cache(key="tracked", prefix="v1", ttl="1d")
+    async def get_tracked(self, db: "Session", requester: "Requester"):
```
nphilipp: … and these type hints should be `AsyncSession`.

fmn/consumer/send_queue.py
```diff
@@ -30,29 +26,31 @@ def giveup_hdlr(details):
 class SendQueue:
     def __init__(self, config):
         self.config = config
+        self._url = get_url_from_config(config)
+        self._url = self._url.update_query(connection_name="FMN consumer to sender")
```
nphilipp: Setting the object member twice looks odd to me. How about …:

```
        _url = get_url_from_config(config)
        self._url = _url.update_query(connection_name="FMN consumer to sender")
```

… or this if it should be more concise:

```
        self._url = get_url_from_config(config).update_query(
            connection_name="FMN consumer to sender"
        )
```

fmn/consumer/send_queue.py
```diff
@@ -30,29 +26,31 @@ def giveup_hdlr(details):
 class SendQueue:
     def __init__(self, config):
```
nphilipp: This is `config: dict`, right?

fmn/core/amqp.py
```diff
@@ -0,0 +1,19 @@
+import ssl
+
+from aio_pika.abc import SSLOptions
+from aio_pika.connection import URL
+
+
+def get_url_from_config(config):
```
nphilipp: This should probably be `config: dict`, too.

fmn/core/config.py
```diff
@@ -7,7 +7,7 @@
 
 
 class CacheModel(BaseModel):
-    backend: str = "dogpile.cache.memory"
+    backend: str = "mem://"
     expiration_time: int = 300
```
nphilipp: Hmm, I could have sworn I used `redis://0.0.0.0` on my end, just to make it clear that this isn’t supposed to live in memory (and set it in tests explicitly to `mem://`). But I guess that got lost in one of the rebases… 🎶

I also have some stuff to e.g. enable configuring TTL differently for Pagure than FASJSON, but I'll put that into my PR once this is through and I have rebased on top.

fmn/database/model/rule.py
```diff
@@ -56,9 +56,10 @@ def select_related(cls) -> Select:
...
 
-    def handle(self, message: "Message", requester: "Requester"):
+    async def handle(self, message: "Message", requester: "Requester"):
         log.debug(f"Rule {self.id} handling message {message.id}")
-        if not self.tracking_rule.matches(message, requester):
+        if not await self.tracking_rule.matches(message, requester):
             return
         for generation_rule in self.generation_rules:
-            yield from generation_rule.handle(message, requester)
+            for notification in generation_rule.handle(message, requester):
```
nphilipp: It looks like this won’t block right now. Any chance (risk) it might do so in the future? I'm asking because on the `TrackingRule` side all this is async…

fmn/database/setup.py
```diff
@@ -16,6 +17,8 @@
 
 def setup_db_schema(engine=None):
     engine = engine or get_sync_engine()
+    if isinstance(engine, Session):
+        engine = engine.get_bind()
```
nphilipp: Let's make this explicit and type-hinted:

```
…
from sqlalchemy.engine import Engine
from sqlalchemy.orm import Session
…
def setup_db_schema(engine_or_session: Engine | Session | None = None) -> None:
    if isinstance(engine_or_session, Session):
        engine = engine_or_session.get_bind()
    else:
        engine = engine_or_session or get_sync_engine()
…
```

fmn/rules/services/distgit.py
```diff
@@ -16,25 +19,30 @@ class DistGitService:
     GROUP_OWNER_LEVELS = ("admin", "commit")
 
     def __init__(self, url):
-        self.url = url
-        self.proxy = PagureSyncProxy(base_url=url)
+        self.url = normalize_url(url)
+        self.proxy = PagureAsyncProxy(base_url=url)
+
+    @cache(ttl="6h", prefix="v1")
+    @handle_http_error(list)
```
nphilipp, line 26: Thinking about it a little too much maybe… but shouldn't this fail if the backend fails rather than papering over it, so the problem can be logged or presented to the user?

fmn/cache/tracked.py
```diff
@@ -0,0 +1,73 @@
...
+    If this happens too frequently, we can just refresh after X minutes have passed and tell
+    users that their changes will take X minutes to be active.
+    """
+
+    async def build(self, db: "Session", requester: "Requester"):
+        log.debug("Building the tracked cache")
+        tracked = Tracked()
+        db_result = await db.execute(Rule.select_related().filter_by(disabled=False))
+        for rule in db_result.scalars():
+            await rule.tracking_rule.prime_cache(tracked, requester)
```
nphilipp, line 50: This will run a potentially huge number of `prime_cache()` calls serially, how long might this take in prod? Maybe we should create tasks for the coroutines and `asyncio.gather()` them, or rather (to limit load on backends) work a limited-size queue of tasks from the coroutines. What do you think?

fmn/rules/services/distgit.py
```diff
@@ -82,22 +93,44 @@ def invalidate_on_message(self, message: "Message"):
...
                     message.body["new_group"],
                     "group",
                 )
         elif message.topic.endswith("pagure.project.group.removed"):
             if message.body["access"] in self.GROUP_OWNER_LEVELS:
-                self._on_owner_changed(
+                await self._on_owner_changed(
                     message.body["project"]["namespace"],
                     message.body["project"]["name"],
                     message.body["new_group"],
```
nphilipp: The `pagure.project.group.removed` message doesn’t have a `new_group` property, but a `removed_groups` (sic!) one. Which is a list of – if I understand the code correctly – up to one string 😂, which is the group we're interested in. I have questions!

fmn/cache/tracked.py
```diff
@@ -0,0 +1,73 @@
...
+    async def get_tracked(self, db: "Session", requester: "Requester"):
+        return await self.build(db=db, requester=requester)
+
+    async def invalidate(self):
+        log.debug("Invalidating the tracked cache")
+        cache_key = list(get_templates_for_func(self.get_tracked))[0]
+        await cache.delete(cache_key)
+
+    async def invalidate_on_message(
+        self, message: "Message", db: "Session", requester: "Requester"
```
nphilipp: … as should this one.

fmn/rules/services/distgit.py
```diff
@@ -82,22 +93,44 @@ def invalidate_on_message(self, message: "Message"):
...
+                "user_or_group": user_or_group,
+            },
+        )
+        # Now rebuild the cache to ease the work on the next incoming message
+        await self.get_owners(namespace, project_name, user_or_group)
+        await self.get_owned(namespace, name, user_or_group)
+        tracked_cache = TrackedCache()
+        await tracked_cache.invalidate()
+        # We can't rebuild the tracked cache here because we don't have access to the DB and
+        # Requester objects.
```
nphilipp: I'm really unsure about the entire concept of "refetch stuff we just invalidated", also about the implementation here:

- We don't know if a message related to the respective project and user/group comes in before the cache entry expires. If so, we cause load elsewhere unnecessarily. Wouldn't this be covered by priming the tracked cache anyway? I can't get rid of the feeling that this might be simpler if we cached closer to the backend service in question. E.g. (see my draft PR #717 😉), the Pagure proxy knows which of its methods deal with say projects/artifacts and can more “surgically” delete cached entries that would yield stale results after a certain change a message is about. Which, if we did it right, would obviate having to throw away and rebuild the whole tracked cache, I think.
- The `get_owners()` and `get_owned()` coroutines are awaited, but serially – we could make them tasks and `asyncio.gather()` them if we think Pagure can handle the load 😅.
- Invalidating the tracked cache without rebuilding it, in conjunction with the order of things in `Consumer.refresh_cache_if_needed()`, would fill the tracked cache and then invalidate it again (actually, invalidate it, the rebuild it, then invalidate it again, if I'm not mistaken).

fmn/rules/services/fasjson.py
```diff
@@ -1,22 +1,40 @@
...
+
+    async def _on_permission_change(self, username):
+        # Don't use the cache.invalidate() decorator as it will invalidate the cache *after* the
+        # function has been run, and we want to refresh it here.
+        await cache.invalidate_func(self.get_user_groups, {"name": username})
+        await self.get_user_groups(username)
+        tracked_cache = TrackedCache()
+        await tracked_cache.invalidate()
+        # We can't rebuild the tracked cache here because we don't have access to the DB and
+        # Requester objects.
```
nphilipp: Essentially, I have the same questions here as above in `distgit.py`.

fmn/rules/services/utils.py
```diff
@@ -0,0 +1,27 @@
...
+                return default_factory()
+
+        return wrapper
+
+    return exception_handler
+
+
+def normalize_url(url):
+    if not url.endswith("/"):
+        url = url + "/"
```
nphilipp: We could cope with URLs that already have a trailing slash:
```
    url = url.rstrip("/") + "/"
```

fmn/rules/tracking_rules.py
```diff
@@ -36,35 +36,23 @@ def __init__(self, *args, **kwargs):
...
-        for container in message.containers:
-            if self.usernames & set(self._requester.get_container_owners(container)):
-                return True
-        for module in message.modules:
-            if self.usernames & set(self._requester.get_module_owners(module)):
-                return True
-        for flatpak in message.flatpaks:
-            if self.usernames & set(self._requester.get_flatpak_owners(flatpak)):
-                return True
+    async def matches(self, message):
```
nphilipp, line 39: Let's have a type hint here – what type of message would this have to be to ensure it has `packages`, `containers`, `modules`, `flatpaks` members?

fmn/rules/tracking_rules.py
```diff
@@ -36,35 +36,23 @@ def __init__(self, *args, **kwargs):
...
-                self._requester.get_owned_by_user(artifact_type="modules", name=username)
-            )
-            cache["flatpaks"].update(
-                self._requester.get_owned_by_user(artifact_type="flatpaks", name=username)
-            )
+            for artifact_type in ArtifactType:
+                owned = await self._requester.distgit.get_owned(
+                    artifact_type.value, username, "user"
+                )
+                getattr(cache, artifact_type.name).update(set(owned))
```
nphilipp, line 55: This looks almost if it would profit from `PagureAsyncProxy.get_projects()` in my PR – it would let us request all artifacts of a user in one go and we could sort them out locally.

fmn/sender/consumer.py
```diff
@@ -14,20 +13,10 @@
 
 class Consumer:
     def __init__(self, config, handler):
-        self._url = URL(config["amqp_url"])
+        self._url = get_url_from_config(config)
         self._destination = config["queue"]
         self._handler = handler
         self._url = self._url.update_query(connection_name=f"FMN sender on {self._destination}")
```
nphilipp: Same comment as above re: setting the `_url` member twice.

fmn/sender/consumer.py
```diff
@@ -14,20 +13,10 @@
...
-        if "tls" in config:
-            self._url = self._url.update_query(auth="EXTERNAL")
-            self._url = self._url.update_query(
-                SSLOptions(
-                    cafile=config["tls"]["ca_cert"],
-                    certfile=config["tls"]["certfile"],
-                    keyfile=config["tls"]["keyfile"],
-                    no_verify_ssl=ssl.CERT_REQUIRED,
-                )
-            )
```
nphilipp, line 30: Is this all obsolete now, does TLS just work without it?

tests/sender/test_consumer.py
```diff
@@ -127,11 +127,11 @@ async def test_with_ssl(mocker):
...
         URL(
             "amqp://rmq.example.com/%2Fvhost?"
-            "connection_name=FMN+sender+on+testdest"
-            "&auth=EXTERNAL"
+            "auth=EXTERNAL"
             "&cafile=/path/to/cacert"
             "&certfile=/path/to/certfile"
             "&keyfile=/path/to/keyfile"
             f"&no_verify_ssl={ssl.CERT_REQUIRED}"
+            "&connection_name=FMN+sender+on+testdest"
```
nphilipp: Not sure about this 100%,but it looks as if we should instantiate the `URL` class of `aio_pika` (well, `yarl`) with a `query={"some": "dict"}` parameter instead of putting it all in one long string (which hopefully shouldn’t care about order of query params in the URL then).

sender.toml.example
```diff
@@ -3,10 +3,10 @@
...
 queue = "email"
 
-[tls]
-ca_cert = "/etc/fedora-messaging/cacert.pem"
-keyfile = "/etc/fedora-messaging/fmn-key.pem"
-certfile = "/etc/fedora-messaging/fmn-cert.pem"
+# [tls]
+# ca_cert = "/etc/fedora-messaging/cacert.pem"
+# keyfile = "/etc/fedora-messaging/sender-key.pem"
+# certfile = "/etc/fedora-messaging/sender-cert.pem"
```
nphilipp, line 9: Same question as elsewhere, don’t we need SSL/TLS configuration any longer? And is the renaming from `fmn-*.pem` to `sender-*.pem` intentional?

nphilipp (changes requested):
Looking good overall! As I mentioned in the chat, some of my comments are less of a request for change as one for discussion (or just stuff which is unclear to me).

fmn/database/model/rule.py
abompard: Good point, I'll make it async

fmn/rules/services/distgit.py
abompard: At that point there is no user, it's the consumer running in the background. That's why I log it and carry on.
A good solution would perhaps be to cache the default result for like 30 seconds? Can Cashews do that?

fmn/cache/tracked.py
abompard: Good idea, let's keep that for a future improvement, no?

fmn/rules/services/distgit.py
abompard: Oh noes...

fmn/rules/services/distgit.py
abompard: Interesting points!

- I'm not sure I understand this one. I'm fine with not rebuilding right after invalidating, because it'll only speed up another non-interactive process, so there's no much point. With regards to the tracked cache, I don't think we can remove parts of it: if I watch the kernel package and you watch it too, when I stopped watching it it should remain in the tracked set.

- Sure, hopefully that won't break Pagure :-)
- If I understand your comment correctly, I don't think that can happen: the messages that `TrackedCache.invalidate_on_message()` reacts to are not the messages that the distgit or fasjson services react to. Or did you mean something else?

fmn/sender/consumer.py
abompard: I moved that to `fmn.core.amqp.get_url_from_config()` because the consumer's consumer (yeah...) use it too.

sender.toml.example
abompard: Yeah we need to be able to do both: in prod it will be a connection to our main broker, and it requires SSL. For development we connect to rabbitmq on tinystage in plain text.

PR #729, opened 2023-02-03
Count the generated notifications and display them in the UI

abompard:
Count the number of notifications generated over the last week, and display it.

It's really not pretty at the moment, maybe @ryanlerch could have a look at it when he gets the time?

Missing: a cleanup script that would remove entries in the new table after some time. This script should be called with Openshift's periodic tasks.

### DIFF
--- a/fmn/core/cli.py
+++ b/fmn/core/cli.py
@@ -1,8 +1,13 @@
+import asyncio
+from datetime import datetime, timedelta
 from importlib.metadata import entry_points
 
 import click
 import click_plugins
+from sqlalchemy import delete, func, select
 
+from ..database.main import async_session_maker, init_async_model
+from ..database.model import Generated
 from . import config
 from .version import __version__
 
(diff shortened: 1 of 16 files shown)


fmn/core/cli.py
```diff
@@ -21,3 +26,31 @@ def cli(settings_file: str | None):
...
+        await init_async_model()
+        async with async_session_maker() as db:
+            result = await db.execute(
+                select(func.count(Generated.id)).filter(Generated.when < limit)
+            )
+            count = result.scalar()
+            if count > 0:
+                click.echo(f"Expiring {count} entries older than {limit}")
+                await db.execute(delete(Generated).filter(Generated.when < limit))
+                await db.commit()
```
nphilipp, line 52: This probably isn’t important, but I guess we could DELETE and count the deleted rows in one go rather than making this two queries, something like (untested):

```
    result = await db.execute(delete(Generated).filter(Generated.when < limit).returning(Generated.id))
    await db.commit()
    if result.rowcount > 0:
        click.echo(f"Expired {result.rowcount} entries older than {limit}")
    else:
        click.echo(f"Nothing to clean up")
```

nphilipp (approved):
Looks good to me! Please merge (as is or with the suggested change, what you prefer).

fmn/core/cli.py
abompard: True, but I kinda like the idea of logging things before they happen, I think it makes it easier to debug. If we were in a performance-critical codepath I would totally adopt your suggestion though :-)

PR #738, opened 2023-02-10
Refactor fasjson and misc fixes and improvements

### LINKED ISSUE #622: Cache requests to FASJSON
Blocked by #714 
~~This is blocked by #621 (this should be implemented in the wrapper)~~

# Story

As a user of the FMN API,
I want that data retrieved from FASJSON is appropriately cached,
so that interacting with the frontend is snappy.

# Acceptance Criteria

- [ ] Allow configuring FASJSON caching lifetime
- [ ] Data retrieved from FASJSON is cached

nphilipp:
Refactor querying FASJSON

This consolidates code accessing FASJSON in FASJSONAsyncProxy which was previously strewn across the API and consumer and adds caching to the proxy methods.

Fixes: #622

Other things:

* Improve handling cashews caches in testing
* Remove synchronous proxies and helpers (which we don’t use anymore)
* Don’t use pytest-flake8 (which is broken with current flake8 and seemingly unmaintained)
* Require cashews >= 5.1.0
* Improve testing wrapper methods
* Remove superfluous database query


tests/backends/test_fasjson.py
```diff
@@ -51,3 +64,106 @@ def test_determine_next_page_params(self, testcase, proxy):
...
+            kwargs = {"self": proxy}
+
+            if ".search_users:" in pattern:
+                func = proxy.search_users
+                kwargs |= {
+                    "username__exact": None if ":username__exact::" in pattern else user,
+                }
+            elif ".get_user:" in pattern:
+                func = proxy.get_user
+                kwargs |= {"username": user}
```
abompard: Hmm, why that and not `kwargs["username"] = user` ? Is it just a matter of style?

nphilipp: I don’t know, it looks weird now you say it. I’ll change it. 😉

nphilipp: Ah, now I remember that it came from the code I copied from Pagure tests which had more than one item to set in the dictionary here (and I was too lazy to write one line for each :joy:).

abompard (approved):

PR #742, opened 2023-02-13
Handle backend errors

abompard:
In case the project does not exist in Pagure, or the user does not exist in FASJSON, we need to handle the HTTP error or the consumer will crash.

Also deal with the service being unavailable (5xx errors) in the same way. The returned default values are not cached.


abompard:
I'd like to get the consumer running in staging, so I'm going to merge this now.

PR #743, opened 2023-02-14
Fix startup crash after the FastAPI update to 0.91.0

abompard:
Since FastAPI 0.91.0, Starlette does not accept adding middleware to an already running application. Change the way we add middleware to fit that.

Ref: https://github.com/tiangolo/fastapi/releases/tag/0.91.0


nphilipp (approved):
LGTM!

PR #744, opened 2023-02-14
Always commit before publishing to Fedora Messaging

abompard:
Otherwise the emitted message may end up in the consumer before we've actually committed to the database.


nphilipp (approved):
LGTM!

PR #746, opened 2023-02-14
Cache the DB request to get all the rules

abompard:
It could be a very big request and it's run on each tracked message. The first changeset caches it for a while, and invalidates the cache when a corresponding message is received.

Also rework the `cache_ttl()` function to use it on the `tracked` and `rules` caches.

### DIFF
--- a/fmn/cache/rules.py
+++ b/fmn/cache/rules.py
@@ -1,0 +1,45 @@
+import logging
+from typing import TYPE_CHECKING
+
+from cashews import cache
+from cashews.formatter import get_templates_for_func
+
+from ..database.model import Rule
+
+if TYPE_CHECKING:
+    from fedora_messaging.message import Message
+    from sqlalchemy.ext.asyncio import AsyncSession
+
+log = logging.getLogger(__name__)
+
+
+class RulesCache:
+    """Cache the rules currently in the database."""
+
+    def __init__(self):
+        self.db: "AsyncSession" | None = None
+
+    @cache.early(
+        key="rules", prefix="v1", ttl="1d", early_ttl="22h"
+    )
+    @cache.locked(key="rules", prefix="v1", ttl="1h")
+    async def get_rules(self):
+        if self.db is None:
+            raise RuntimeError("You must use set the db attribute first.")
+        log.debug("Building the rules cache")
+        result = await self.db.execute(Rule.select_related().filter_by(disabled=False))
+        log.debug("Built the rules cache")
+        return list(result.scalars())
+
+    async def invalidate(self):
+        log.debug("Invalidating the rules cache")
+        cache_key = list(get_templates_for_func(self.get_rules))[0]
+        await cache.delete(cache_key)
+
+    async def invalidate_on_message(self, message: "Message"):
+        if (
+            message.topic.endswith("fmn.rule.create.v1")
+            or message.topic.endswith("fmn.rule.update.v1")
+            or message.topic.endswith("fmn.rule.delete.v1")
+        ):
+            await self.invalidate()
--- a/fmn/cache/rules.py
+++ b/fmn/cache/rules.py
@@ -5,6 +5,7 @@ from cashews import cache
 from cashews.formatter import get_templates_for_func
 
 from ..database.model import Rule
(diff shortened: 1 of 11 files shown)


fmn/cache/rules.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class RulesCache:
+    """Cache the rules currently in the database."""
+
+    def __init__(self):
+        self._db: "AsyncSession" | None = None
+
+    def set_db(self, db: "AsyncSession"):
+        self._db = db
```
nphilipp: Why not simply manipulate/use the member directly here (and call it `self.db`)?

fmn/cache/rules.py
```diff
@@ -0,0 +1,52 @@
...
+    @cache.early(
+        key="rules", prefix="v1", ttl=cache_ttl("rules"), early_ttl=cache_early_ttl("rules")
+    )
+    async def get_rules(self):
+        return await self.build()
+
+    async def invalidate(self):
+        log.debug("Invalidating the rules cache")
+        cache_key = list(get_templates_for_func(self.get_rules))[0]
+        await cache.delete(cache_key)
```
nphilipp: I’m not sure why there can be more than one template for a function, but looking at `get_templates_for_func()`, it seems possible and we could cope with that like this:

```
cache_keys = get_templates_for_func(self.get_rules)
await asyncio.gather(*(cache.delete(key) for key in cache_keys))
```

fmn/cache/rules.py
```diff
@@ -0,0 +1,52 @@
...
+        log.debug("Building the rules cache")
+        result = await self._db.execute(Rule.select_related().filter_by(disabled=False))
+        log.debug("Built the rules cache")
+        return list(result.scalars())
+
+    @cache.early(
+        key="rules", prefix="v1", ttl=cache_ttl("rules"), early_ttl=cache_early_ttl("rules")
+    )
+    async def get_rules(self):
+        return await self.build()
```
nphilipp: Do we use the `build()` method directly? Because we could apply both decorators to the same method (or use `lock=True` on just the `@cache.early()` decorator) according to the [Cashews docs](https://github.com/Krukov/cashews/#locked).

fmn/cache/tracked.py
```diff
@@ -43,22 +43,27 @@ class TrackedCache:
...
         return tracked
 
-    @cache.early(key="tracked", prefix="v1", ttl="1d", early_ttl="22h")
-    async def get_tracked(self, db: "AsyncSession", requester: "Requester"):
-        return await self.build(db=db, requester=requester)
+    @cache.early(
+        key="tracked", prefix="v1", ttl=cache_ttl("tracked"), early_ttl=cache_early_ttl("tracked")
+    )
+    async def get_tracked(self):
+        return await self.build()
```
nphilipp: Same question as with `RulesCache.build()`, `RulesCache.get_rules()` I guess 😉

nphilipp (changes requested):

fmn/cache/rules.py
abompard: Yeah, I wasn't sure. I suppose I could do that as well. It feels more explicit and less hackish to go through a function call instead of manipulating the attribute directly, but maybe that's just me.
I'll change it to `.db`

fmn/cache/rules.py
abompard: Yeah I don't see why we would have more than one template either. But indeed your solution is better.

fmn/cache/rules.py
nphilipp: I imagine it’s idiomatic in Javascript e.a. to use accessor methods across the board? Python looks like an outlier here, with the convention to access members directly and if you need to intercept setting/getting, then turn it into a property.

fmn/cache/rules.py
abompard: Yeah, I might merge those two functions. It was practical for unit tests, but I suppose I can do without. I'll keep a second decorator though, otherwise the lock lifetime does not have a TTL.

nphilipp (approved):
LGTM!

PR #748, opened 2023-02-16
Add Admin Interface

### LINKED ISSUE #756: Add a way for sysadmins to disable/re-enable rules
# Story

As a Fedora administrator,
I want to disable and enable rules of arbitrary users,
so that I can react to them causing problems.

# Acceptance Criteria

- [ ] User groups can be specified for administrative access in FMN configuration
- Such admins can
  - [ ] Delete rules by their ID (mentioned in log files)
  - [ ] List disabled rules
  - [ ] Re-enable such disabled rules

ryanlerch:


nphilipp:
The latest push configures the frontend to request group info from Ipsilon, in the API pulls user info from Ipsilon, makes it available as `.user_info` in the API identity structure and updates most tests to follow suit. Additionally, fixes a couple minor things which partly need squashing with earlier commits (the commit without signed-off-by line).

WIP, because this isn’t used anywhere and one test(\*) fails and I couldn’t find out why. 😉

(\*): tests/api/test_handlers.py::TestUserHandler::test_create_user_rule[success]

frontend/src/components/AdminLink.vue
```diff
@@ -0,0 +1,29 @@
...
+const { isLoading, isError, data, error } = useQuery(
+  url,
+  apiGet as QueryFunction<boolean>,
+  {
+    retry: false,
+  }
+);
+/*eslint-enable */
+
+const isAdmin = computed(() => data );
```
abompard: You could also do `const { isLoading, isError, data: isAdmin, error } = useQuery(` and this way you don't need to define `isAdmin` as a derivative of `data`.

frontend/src/components/AdminLink.vue
```diff
@@ -0,0 +1,29 @@
...
+import { computed } from "vue";
+import type { QueryFunction } from "react-query/types/core";
+import { useQuery } from "vue-query";
+import { useUserStore } from "../stores/user";
+
+const userStore = useUserStore();
+
+const url = `/api/v1/admin`;
+
+/*eslint-disable */
```
abompard: Why disable eslint?

frontend/src/components/AdminLink.vue
```diff
@@ -0,0 +1,29 @@
...
+    retry: false,
+  }
+);
+/*eslint-enable */
+
+const isAdmin = computed(() => data );
+</script>
+
+<template>
+  <router-link v-if="isAdmin && userStore.loggedIn" to="/admin" class="nav-link"
```
abompard: Can `isAdmin` be true if the user is not logged in?

fmn/api/handlers/admin.py
```diff
@@ -0,0 +1,68 @@
...
+    else:
+        return False
+
+
+@router.get("/rules", response_model=list[api_models.Rule], tags=["users/rules"])
+async def get_disabled_rules(
+    identity: Identity = Depends(get_identity_optional),
+    db_session: AsyncSession = Depends(gen_db_session),
+):
+    if not identity.admin:
```
abompard: Instead of repeating this test in all the endpoints, I would suggest adding this to `fmn/api/auth.py`:

```py
async def get_identity_admin(identity: Identity = Depends(get_identity)):
    if not identity.admin:
        raise HTTPException(status.HTTP_403_FORBIDDEN, detail="Not an admin")
    return identity
```

And using this function as a dependency instead of `get_identity_optional`.

abompard: I've implemented this suggestion in [95cfbfb](https://github.com/fedora-infra/fmn/pull/748/commits/95cfbfbc6f3b8be55fe3a5dd92cf8105a2971f41).

tests/api/handlers/test_admin.py
```diff
@@ -11,3 +16,104 @@ def test_admin_rules_forbidden(self, client, api_identity):
...
+                        "severities": [],
+                        "topic": None,
+                        "my_actions": False,
+                    },
+                }
+            ],
+            "generated_last_week": 0,
+        }
+
+    @mock.patch("fmn.api.handlers.admin.publish")
```
abompard, line 57: You can use `fedora_messaging.testing.mock_sends()` instead, see other tests for a usage example.

tests/api/handlers/test_admin.py
```diff
@@ -11,3 +16,104 @@ def test_admin_rules_forbidden(self, client, api_identity):
...
+        edited_rule.disabled = False
+        success_message = RuleUpdateV1(
+            body={
+                "rule": edited_rule.dict(),
+                "user": api_models.User.from_orm(db_rule_disabled.user).dict(),
+            }
+        )
+        publish.assert_awaited_once_with(success_message)
+
+    @mock.patch("fmn.api.handlers.admin.publish")
```
abompard, line 88: Same here (about `fedora_messaging.testing.mock_sends()`)

ryanlerch: oh, i was copying something already in place in handlers/test_users.py

ryanlerch: that said, still changing it over to fedora_messaging.testing.mock_sends()

fmn/api/handlers/users.py
```diff
@@ -22,6 +22,16 @@
...
 
 
+@router.post("/me", response_model=api_models.User, tags=["users"])
+async def post_me(
+    identity: Identity = Depends(get_identity),
+    db_session: AsyncSession = Depends(gen_db_session),
+):
+    user = await User.async_get_or_create(db_session, name=identity.name)
+    user.is_admin = identity.admin
+    return user
```
nphilipp: Why POST?

fmn/api/handlers/admin.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+@router.patch("/rules/{id}/disabled", response_model=api_models.Rule, tags=["users/rules"])
+async def edit_rule_disabled(
+    id: int,
+    disable: api_models.RuleDisable,
+    identity: Identity = Depends(get_identity_admin),
+    db_session: AsyncSession = Depends(gen_db_session),
+):
+    rule_db = (await db_session.execute(Rule.select_related().filter(Rule.id == id))).scalar()
```
nphilipp: The route/method looks odd in a restful API, I’d rather have it be sth. like this:

```
@router.patch("/rules/{id}", response_model=api_models.Rule, tags=["users/rules"])
async def patch_rule(
    id: int,
    rule_patch: RulePatch,  # like Rule, only all optional
    identity ...,
    db_session ...,
):
    ...
    if rule_patch.disabled is not None:
        rule_db.disabled = rule_patch.disabled
    ...
```

nphilipp (changes requested):

nphilipp:
WIP: Make the admin/rules endpoints more restful

- Make the endpoint paths about “the things”, not the operations.
- Don’t include the rule ID in the API model for patching, it’s in the
  path already.  It’s still there frontend-side, but that’s because I
  don’t know what I’m doing… (but it really shouldn’t be there in the
  backend).

WIP, because I'm just cargo-culting the frontend changes and I’m sure
they don’t work together as-is with the backend because the frontend
still has the ID while the backend doesn’t. 😅

nphilipp (approved):

PR #762, opened 2023-02-27
Stop ignoring whole hierarchies in tests

nphilipp:


abompard:
Oh ! Yeah that's pretty bad :grimacing:

nphilipp:
Isn’t it? 😁

PR #764, opened 2023-02-28
Use Ruff instead of flake8 & isort

abompard:
I just discovered [Ruff](https://beta.ruff.rs/docs/) because Bodhi is using it. It's pretty cool!

For VS Code support, just install [the Ruff extension](https://marketplace.visualstudio.com/items?itemName=charliermarsh.ruff) and disable the flake8 extension.


fmn/api/main.py
```diff
@@ -16,7 +16,7 @@
 
 tags_metadata = [
     {"name": "users", "description": "Work with users"},
-    {"name": "users/rules", "description": "Work with users’ rules"},
+    {"name": "users/rules", "description": "Work with users' rules"},
```
nphilipp: This is wrong, the original string contains a typographical apostrophe (U+2019) which shouldn’t be replaced by its ASCII counterpart (U+0027).

fmn/api/messaging.py
```diff
@@ -30,6 +30,14 @@ def _publish(message):
...
+
 async def publish(message):
     # Fire and forget
-    asyncio.ensure_future(run_in_threadpool(_publish, message=message))
+    # https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
+    task = asyncio.create_task(run_in_threadpool(_publish, message=message))
+    _background_tasks.add(task)
+    # To prevent keeping references to finished tasks forever, make each task
+    # remove its own reference from the set after completion:
+    task.add_done_callback(_background_tasks.discard)
```
nphilipp, line 43: This looks unrelated to this PR…

fmn/backends/base.py
```diff
@@ -53,7 +53,6 @@ def base_url_with_trailing_slash(self) -> str:
     def api_url(self) -> str | None:
         return self.base_url
 
-    @ft_cache
     def __str__(self) -> str:
```
nphilipp, line 56: This must be the potential memory leak 😉, only it shouldn’t be…

The reason why I cached `APIClient.__str__()` was to speed up cache key generation for `cashews` and I’d like to retain that optimization of possible (it speeds up `str()` by a factor of about 3). And I think it is, we could do either or both of:

* Make the proxy objects singletons based off the arguments passed to the class when constructing (and `Requester` as well while we’re at it) ⇒ Only one return value cached per singleton/class.
* Clear the cache on the method when finalizing in `__del__()`.

fmn/backends/pagure.py
```diff
@@ -32,7 +32,7 @@ class PagureRole(IntFlag):
     GROUP_ROLES = GROUP_ROLES_MAINTAINER | TICKET
 
 
-# Python < 3.11 doesn’t allow iterating over combined flag values
+# Python < 3.11 doesn't allow iterating over combined flag values
```
nphilipp: Same typographical vs. ASCII apostrophe thing.

fmn/cache/util.py
```diff
@@ -25,7 +26,7 @@ def cache_arg(arg: str, scope: str | None = None) -> Callable[[str, str | None],
 
     @ft_cache
     def get_cache_arg(*cached_fn_args, **cached_fn_kwargs) -> Any:
-        f"""Get the cache arg {arg} for scope {scope}."""
+        """Get the cache arg {arg} for scope {scope}."""
```
nphilipp: Hmm, we should probably remove this… the return value is a function but it is directly passed onto cashews which probably doesn’t need the docstring.

fmn/database/migrations/env.py
```diff
@@ -1,7 +1,7 @@
 from alembic import context
 from sqlalchemy import engine_from_config, pool
 
-# Alembic imports this module in a way that it can’t do relative imports from outside the
+# Alembic imports this module in a way that it can't do relative imports from outside the
```
nphilipp: Again… :wink:

tests/sender/test_irc.py
```diff
@@ -17,7 +17,7 @@ async def _send_welcome():
         await asyncio.sleep(0.5)
         handler._client._dispatcher(transport, Event("welcome", "server", "user"))
 
-    asyncio.create_task(_send_welcome())
+    asyncio.create_task(_send_welcome())  # noqa: RUF006
```
nphilipp, line 22: Maybe add a comment what RUF006 is about?

tests/sender/test_matrix.py
```diff
@@ -17,7 +17,7 @@ async def test_matrix_connect(mocker):
     await handler.setup()
 
     assert client.user_id == "fmnuser"
-    assert client.access_token == "dummytoken"
+    assert client.access_token == "dummytoken"  # noqa: S105
```
nphilipp, line 21: Same here.

nphilipp (changes requested):

fmn/api/main.py
abompard: Ah, this is part of a rule that aims at avoiding confusable characters. I don't have any opinion on this, so if you prefer the typographical apostrophe I'm happy to add it to the allowlist.

fmn/api/messaging.py
abompard: It actually is, it's an issue discovered by the "bugbear" plugin (or rather its reimplementation in ruff). As stated in `create_task()`'s doc linked above, it's necessary to hold a reference to future/tasks created by `ensure_future` and `create_task`, otherwise the task may be garbage-collected before it has had a chance to run. They recommend the trick of adding the reference to a list and removing it on completion.

fmn/backends/base.py
abompard: Or we could store the computed value in a local class variable such as `_str` and just return that when `__str__()` is called. I'd prefer this or the cache clear on `__del__()`, any preference between the two?

(I kinda dislike singletons as they can make testing harder)

fmn/cache/util.py
abompard: The rule that picked it up said that the f-string mark would prevent Python from recognizing it as a docstring. Aside from that I'm fine with keeping it as a plain string if it's useful doc for humans, or removing it if not.

tests/sender/test_irc.py
abompard: Yep, will do. It's the issue about not storing references to tasks: https://beta.ruff.rs/docs/rules/asyncio-dangling-task/

fmn/api/main.py
nphilipp: This looks to be caused by one of `RUF001`, `RUF002`, `RUF003`. I think we should configure these rules to be ignored.

fmn/api/messaging.py
nphilipp: 👍

fmn/backends/base.py
nphilipp: Good idea, something like this should do the trick:

```
def __str__(self) -> str:
    try:
        return self._str
    except AttributeError:
        clsname = type(self).__name__
        self._str = f"{clsname}({self.base_url!r})"
        return self._str
```

fmn/cache/util.py
nphilipp: > The rule that picked it up said that the f-string mark would prevent Python from recognizing it as a docstring.

Ah, wasn’t aware of that. It was supposed to contain the actual values for `arg` and `scope`, as the function is dynamically generated. I’d say, just remove it.

fmn/backends/base.py
```diff
@@ -53,10 +53,11 @@ def base_url_with_trailing_slash(self) -> str:
...
-        clsname = type(self).__name__
-        return f"{clsname}({self.base_url!r})"
+        if not hasattr(self, "_str"):
+            clsname = type(self).__name__
+            self._str = f"{clsname}({self.base_url!r})"
+        return self._str
```
nphilipp: [I benchmarked it](https://gist.github.com/nphilipp/f211c8a016a8057e7f4b89f6b5530714) and try/except is about 30% faster than manual checks (`foo` is try/except, the others are variants of conditionals):

```nils@makake:~/test/python> ./try_except_vs_cond.py 
foo: 0.0718677620170638
bar: 0.10609414198552258
baz: 0.10942566799349152
nils@makake:~/test/python>
```

abompard: Interesting! I was just trying to avoid the duplicate `return` calls. I'll go back to your version, thanks!

nphilipp: Heh, talked about this with a friend and he pointed me to a flaw in how I benchmarked this: It assumes long-lived objects, but we don’t have that with `gen_requester()` and `get_last_messages()` from `fmn.api.handlers.utils`. For short-lived objects, the performance reverses: try/except is a bit more than twice as expensive as using conditionals (gist [here](https://gist.github.com/nphilipp/6c4400f97eac7258568e38fb939d84ab)):

```
nils@makake:~/test/python> ./try_except_vs_cond2.py 
foo: 0.8480359880195465
bar: 0.3848317340016365
baz: 0.3872268960112706
nils@makake:~/test/python>
```

I lean towards making these cache their `Requester` and `DatagrepperAsyncProxy` objects, respectively. What do you think?

abompard: Hmm yeah I suppose they can live for the duration of the application.

nphilipp (changes requested):
Hmm:

```
…
lint: install_deps> python -I -m pip install ruff
lint: freeze> python -m pip freeze --all
lint: pip==23.0.1,ruff==0.0.254,setuptools==67.4.0,wheel==0.38.4
lint: commands[0]> ruff check fmn/ tests/
tests/database/model/base.py:23:16: UP038 [*] Use `X | Y` in `isinstance` call instead of `(X, Y)`
tests/database/model/base.py:29:16: UP038 [*] Use `X | Y` in `isinstance` call instead of `(X, Y)`
tests/database/model/base.py:46:16: UP038 [*] Use `X | Y` in `isinstance` call instead of `(X, Y)`
tests/database/model/base.py:52:16: UP038 [*] Use `X | Y` in `isinstance` call instead of `(X, Y)`
Found 4 errors.
[*] 4 potentially fixable with the --fix option.
lint: exit 1 (0.05 seconds) /__w/fmn/fmn> ruff check fmn/ tests/ pid=3176
  py310: OK (88.81=setup[1.17]+cmd[9.54,55.78,22.31] seconds)
  py311: OK (98.03=setup[1.24]+cmd[16.86,56.76,23.16] seconds)
  black: OK (3.59=setup[1.02]+cmd[1.43,1.13] seconds)
  lint: FAIL code 1 (2.60=setup[2.54]+cmd[0.05] seconds)
  evaluation failed :( (193.08 seconds)
…
```

Not sure if I consider this broken, we’ve used `isinstance(…, (TypeA, TypeB))` for ages and the new method `isinstance(…, TypeA | TypeB)` is only mentioned as an afterthought in the [docs](https://docs.python.org/3/library/functions.html#isinstance). What do you think?

nphilipp (approved):
LGTM!

PR #765, opened 2023-02-28
change "irc client" to "chat client"

mattdm:
So that if we remove irc, or matrix morphs to something else, or whatever, we don't have to update


nphilipp (approved):
Looks good to me!

nphilipp:
The failing backend CI tests are unrelated to this change.

PR #777, opened 2023-03-06
Consumer: get a session from the sessionmaker on each incoming message

abompard:
Otherwise we get this kind of exception:
```
sqlalchemy.exc.InterfaceError: (sqlalchemy.dialects.postgresql.asyncpg.InterfaceError) <class 'asyncpg.exceptions._base.InterfaceError'>: cannot perform operation: another operation is in progress
```

The fact that the sessionmaker has a connection pool should solve the potential performance issue.

### DIFF
--- a/fmn/consumer/consumer.py
+++ b/fmn/consumer/consumer.py
@@ -6,6 +6,7 @@ from aio_pika.exceptions import AMQPConnectionError
 from fedora_messaging import message
 from fedora_messaging.config import conf as fm_config
 from fedora_messaging.exceptions import Nack
+from sqlalchemy.ext.asyncio import AsyncSession
 
 from ..cache import configure_cache
 from ..cache.rules import RulesCache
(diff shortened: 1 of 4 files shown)


fmn/consumer/consumer.py
```diff
@@ -55,13 +53,17 @@ def __call__(self, message: message.Message):
...
+            # Process message
+            try:
+                await self.handle(message)
+            except Exception:
+                await db.rollback()
+                raise
+            else:
+                await db.commit()
 
     async def handle(self, message: message.Message):
```
nphilipp: Hmm, I don’t like storing ephemeral objects in object members to pass them around, we should rather pass it into the method explicitly (if needs be). I can think of two alternative ways:

* It doesn’t look as if `Consumer.handle()` is used outside of the class, so how about we mark it as not public (e.g. rename to `_handle()`) and pass in the session object explicitly.
* We could merge both methods into `handle` and make whether or not the transaction should be rolled back on error a parameter.

nphilipp (changes requested):

fmn/consumer/consumer.py
abompard: Yeah, I'll modify handle, it seems better indeed.

nphilipp (approved):
LGTM!

PR #780, opened 2023-03-06
Don’t query Datagrepper anymore

### LINKED ISSUE #776: Don’t query datagrepper
# Stories

As a user of FMN,
I want that API requests return in a short time,
so that interactive use feels snappy.

As a developer of FMN,
I want that it doesn’t contain half-implemented, abysmally performing features,
so that I don’t have to cry myself to sleep.

# Acceptance Criteria

- [ ] FMN doesn’t query datagrepper in bulk

# Background

Originally, we wanted to give people an idea how many notifications a rule they’re about to add would produce in, say, a week. However, even if we cached a full week’s worth of bus messages, processing them takes ages because of their sheer amount. Thus, all users would normally see about this feature is a spinner while the query is still underway, making it less than useful.

nphilipp:
We had this initially to generate previews for new or edited rules, but it proved to costly.

Fixes: #776


tests/api/handlers/test_utils.py
```diff
@@ -0,0 +1,8 @@
+from fmn.api.handlers import utils
+from fmn.rules.requester import Requester
+
+
+def test_gen_requester():
```
abompard: I don't think anyone uses the `gen_requester()` function anymore, it can be removed as well.

nphilipp: Good catch!

nphilipp: In fact, `get_message()` below can go away, too.

abompard (approved):

PR #791, opened 2023-03-07
Bring back flake8 configuration

nphilipp:
This was removed when we switched to using Ruff instead of a collection of individual linter and static analysis tools.

The reason to bring it back is that flake8 doesn’t complain about overly long lines, if it’s configured to be run e.g. by the editor or a global pre-commit hook.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -40,3 +40,9 @@ commands=
     rm -rf _build
     rm -rf _source
     sphinx-build -W -b html -d {envtmpdir}/doctrees .  _build/html
+
+# We use Ruff instead of flake8 but configure it appropriately so it doesn’t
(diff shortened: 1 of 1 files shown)


abompard:
Hey! Have you tried to configure your editor to run Ruff instead? VSCode has a pretty good ruff plugin! :-)
I haven't looked at configuration for global pre-commit hooks though

nphilipp:
I’ve thought about it, but I have so many repos which use flake8 I can’t switch it off wholesale. Not sure if there’s a way to tell VIM to use one set of "write file hooks" for one repo and another for others…

abompard (approved):
Alright, but let's remember to keep them in sync then :-)

PR #798, opened 2023-03-08
Add Sphinx documentation

abompard:
Add a Sphinx skeleton configured to understand Markdown.


abompard:
FYI @nphilipp  :-)

abompard:
Merging because it's just adding stuff and not likely to break existing stuff.

Issue #788, opened 2023-03-07
The "rules count" does not take into account the rule filter


abompard:
Please keep the "rules quantity" component dynamic so that it falls in line with the number of rules being shown.
Either that or call it "5 total rule(s)"
The number of rules shown in the screenshot below is definitely not 5.

![image](https://user-images.githubusercontent.com/95780/223460125-f39bad78-0095-4422-b269-95c139b74f65.png)

PR #805, opened 2023-03-14
Tweak rule lists, edit and create rules forms

### LINKED ISSUE #747: Suggest a better name for new rules

### LINKED ISSUE #784: Disallow rules with an empty name
Rules should have a name, otherwise users can't click on it to edit or delete them.

Make sure you change both the API's Pydantic model and the UI's validation tags.

### LINKED ISSUE #785: Rules tracking Artifacts don't remember their artifacts in the UI
When a rules is set to track an RPM artifact, the field "Artifact names" is empty when going back to edit the rule.

This is probably linked to the Multiselect widget being very picky with default values when it's getting the data from an API call.

### LINKED ISSUE #787: Should we allow a user's rules to have the same name?
It would force users to pick relevant names, but it could also be a bit annoying. What do you think?

### LINKED ISSUE #789: The placement of the New Rule button can be misleading
Also please rename the "Add a new rule" button to "Add new rule" and relocate it under the "My Rules" header.
This is to avoid the probable confusion that might happen when these two are put side-by-side.

image

For a newbie (who is using the service for the first time), it might look like clicking on that button would lead to the user creating a rule for "Specific Artifacts".
It would, of course, grow on them when they see the list changing as they change the contents of the drop down box.
But initially when the list is empty, I am 100% sure people would get confused as to what exactly the button is intended to do.

ryanlerch:
Makes a bunch of changes to rules lists and the create / edit rules forms:

* make the rule title less important, i.e. optional, so when a rule title is not added, we show a generated one now in the my rules list.
* The my rules list now works better on smaller screens, and displays the information a little more cleaner and succinctly
* The rule title field is now not at the top of the create / edit rule form
* The add new rule button is no longer in the list of rules, but up the top of the my rules page.
* the my rules page filter is now to the right where the add new rule button was, and is now using the multiselect field rather than an ordinary form
* the total number of rules is now listed in a badge at the top of the page near the "My Rules" header

Resolves: #789
Resolves: #787
Resolves: #784
Resolves: #785
Resolves: #747


abompard (changes requested):
Looks much better, but if I click on the rule title field, change it, and press "Enter", the rule gets deleted.

ryanlerch:
> Looks much better, but if I click on the rule title field, change it, and press "Enter", the rule gets deleted.

that's not ideal. looking into it now

ryanlerch:
> > Looks much better, but if I click on the rule title field, change it, and press "Enter", the rule gets deleted.
> 
> that's not ideal. looking into it now

Fixed now -- adding a "type='button'" to the delete CButton fixes the issue. we already had a type=submit on the edit button, but apparently because the delete button was first, we need the additional type=button to not default to it on enter

abompard:
Some comments

- This does not seem to solve #788
- The "Rule enabled" toggle ends up below the list of tracked artifacts when one of the corresponding tracking rules are selected. If there are many artifacts, I think the toggle may be hardly visible
- The filter tags are displayed even if the filter is empty, they should probably only be shown if the user has actually selected a filter
Capture d’écran du 2023-03-14 16-08-29

ryanlerch:
> Some comments
> 
> * This does not seem to solve [The "rules count" does not take into account the rule filter #788](https://github.com/fedora-infra/fmn/issues/788)

Removed this issue from this PR for now -- got an idea on how to fix, but will do a seperate PR

> * The "Rule enabled" toggle ends up below the list of tracked artifacts when one of the corresponding tracking rules are selected. If there are many artifacts, I think the toggle may be hardly visible

Moved this up to the top of the list -- however going to work on another PR to make this cleaner -- its just a bigger task than this PR needs

> * The filter tags are displayed even if the filter is empty, they should probably only be shown if the user has actually selected a filter
>   Capture d’écran du 2023-03-14 16-08-29

This is actually an pre-existing issue -- it only shows when creating a new generation rule -- if you refresh they go away. will file a seperate issue for this one

PR #809, opened 2023-03-15
Don't show empty genrule filter labels when adding a new genrule

### LINKED ISSUE #808: Generation rule empty params showing
After creating a new generation rule, the empty parameter labels show up like:

Capture d’écran du 2023-03-14 16-08-29

Note that this only shows up when adding a new generation rule. once you save it and reload the page, the empty parameter labels arent shown (which is what is expected)

ryanlerch:
Resolves: #808

### DIFF
--- a/frontend/src/components/rule-edit/generation-rule/GenerationRuleListItem.vue
+++ b/frontend/src/components/rule-edit/generation-rule/GenerationRuleListItem.vue
@@ -39,9 +39,7 @@ const handleDeleteClicked = () => {
           <span
             class="badge bg-light border border-info text-secondary fw-normal me-1"
             v-if="
-              f_params !== null &&
-              f_params !== false &&
-              !(Array.isArray(f_params) && f_params.length === 0)
+              !!f_params && !(Array.isArray(f_params) && f_params.length === 0)
             "
             ><strong>{{ f_name }}</strong>
             <template v-if="typeof f_params !== 'boolean'"> : </template>


frontend/src/components/rule-edit/generation-rule/GenerationRuleListItem.vue
```diff
@@ -41,6 +41,7 @@ const handleDeleteClicked = () => {
             v-if="
               f_params !== null &&
               f_params !== false &&
+              f_params !== undefined &&
```
abompard: Two things:
- testing for a variable being undefined in JS is usually done with `typeof variable == "undefined"`, I'm not sure why it is better than comparing to `undefined` directly like you do but maybe there are corner cases
- it feels like this whole test could be summed up as testing whether `f_params` is *truthy*, in JS there are several ways of converting a variable to a boolean, the one I've seen most is `!!variable` (two exclamation points). Would that work? I think we'd still need to test for empty array, but the three others should be covered.

ryanlerch: !!variable works perfectly. thanks

frontend/src/components/rule-edit/generation-rule/GenerationRuleListItem.vue
```diff
@@ -32,16 +32,15 @@ const handleDeleteClicked = () => {
         :key="`${dest.protocol}:${dest.address}`"
       />
       <div>
+        {{ props.rule.filters }}
```
abompard: Looks like a forgotten debugging line

abompard (changes requested):

PR #810, opened 2023-03-15
Add a filtered count at the top of the rules list

### LINKED ISSUE #788: The "rules count" does not take into account the rule filter

ryanlerch:
Resolves: #788

### DIFF
--- a/frontend/src/components/RulesList.vue
+++ b/frontend/src/components/RulesList.vue
@@ -31,22 +31,23 @@ const filteringOptions = computed(() => {
 });
 
 const rules = computed(() => {
-  return [...props.rules].sort(
-    (a, b) =>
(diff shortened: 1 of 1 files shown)


abompard:
It feels a bit redundant when no filters are selected, with the badge and the table label displaying the same number. How about something like "4 rules, 2 shown" when a filter is selected and just "4 rules" otherwise?

ryanlerch:
and remove the badge?

ryanlerch:
badge removed. when no filter it reads "6 rules" when a filter it reads "6 rules, 2 shown"

abompard (approved):

PR #819, opened 2023-03-24
Add a command to refresh the cache before it expires

abompard:
This should allow us to avoid blocking the consumer(s) if the cached data becomes too old.

The command should be run in a cronjob, I'd say something like every hour. If the data is not stale, it will not do anything.


fmn/core/config.py
```diff
@@ -12,14 +12,15 @@
 class CacheArgsModel(BaseModel):
     ttl: CashewsTTLTypes | None = None
     lock_ttl: CashewsTTLTypes | None = None
+    early_ttl: CashewsTTLTypes | None = None
 
-    class Config:
-        extra = "allow"
+    # class Config:
+    #     extra = "allow"
```
nphilipp: Remove this instead of commenting it out?

nphilipp (changes requested):

fmn/core/config.py
abompard: Indeed!

nphilipp (approved):
LGTM!

PR #830, opened 2023-03-30
Fix the artifact selection widget

### LINKED ISSUE #829: Editing rule doesn’t load artifacts
When editing a rule about specific artifacts, the “Artifact Names" field is left empty, it should be filled with the artifacts named when creating the rule.

abompard:
It now handles default values and does the sorting into categories (the API used to do that).

Fixes: #829

### DIFF
--- a/fmn/api/handlers/misc.py
+++ b/fmn/api/handlers/misc.py
@@ -5,7 +5,6 @@ from fastapi import APIRouter, Depends, HTTPException, Query
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from ...backends import PagureAsyncProxy, get_distgit_proxy
-from ...core.constants import ArtifactType
 from ...database.migrations.main import alembic_migration
 from .. import api_models
 from ..database import gen_db_session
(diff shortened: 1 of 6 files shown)


fmn/api/handlers/misc.py
```diff
@@ -57,40 +57,18 @@ async def get_owned_artifacts(
...
-                    }
-                )
-    return artifacts
+        result.append(
+            {
+                "type": project["namespace"],
+                "name": project["name"],
+            }
+        )
+    return result
```
nphilipp, line 70: How about using a list comprehension instead of appending piece-meal? Would make it simpler and probably more efficient. E.g.: 

```
    ...
    projects = [
        {"type": project["namespace"], "name": project["name"]}
        for project in await distgit_proxy.get_projects(pattern=f"*{name}*")
    ]
    return projects
    ...
```

Also, the test data on src.tinystage.test uses the (wrong) namespace `rpm` instead of `rpms` which makes the API throw a fit on validation. I think we should filter for validity, so FMN copes with either “garbage” on src.fp.o or a new namespace being introduced. Unfortunately, we can’t simply do `project["namespace"] in ArtifactType`, so it needs a little more work (which I’ll tackle in #831).

nphilipp (approved):
Looks good to me! #829 seems to be fixed with it (in Vagrant).

PR #833, opened 2023-03-30
Filter unexpected project namespaces from Pagure

### LINKED ISSUE #831: Don’t trust that Pagure namespaces are valid `ArtifactTypes`
# Story

As a user of FMN,
I want that it filters unexpected input from services it depends on,
so that it copes well instead of throwing errors.

# Acceptance Criteria

- [x] FMN filters namespaces received from Pagure and only processes those it considers valid ArtifactTypes.

# Background

The distgit/Pagure instance in tinystage mistakenly uses `rpm` as the namespace for RPM packages (instead of `rpms`). This makes output validation of `/artifacts` throw a fit.

nphilipp:
The API endpoints proxying Pagure validate their results but querying Pagure could yield project namespaces not known to FMN (by mistake or new kinds of artifacts introduced) which makes the handler croak.

This was found to be necessary because of a bug in tinystage, where RPM package projects in test data used the `rpm` namespace instead of `rpms`. See fedora-infra/tiny-stage#44.

Fixes: #831

### DIFF
--- a/fmn/core/constants.py
+++ b/fmn/core/constants.py
@@ -1,4 +1,5 @@
 from enum import Enum
+from functools import cache
 
 
 class ArtifactType(Enum):
@@ -7,6 +8,15 @@ class ArtifactType(Enum):
     containers = "containers"
     modules = "modules"
     flatpaks = "flatpaks"
+
+    @classmethod
+    @cache
+    def has_value(cls, value):
+        try:
+            cls(value)
+        except ValueError:
+            return False
+        return True
 
 
 DEFAULT_MATRIX_DOMAIN = "fedora.im"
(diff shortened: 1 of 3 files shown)


fmn/core/constants.py
```diff
@@ -1,7 +1,18 @@
-from enum import Enum
+from enum import Enum, EnumMeta
+from functools import cache
 
 
-class ArtifactType(Enum):
+class ContainsEnumMeta(EnumMeta):
+    @cache  # noqa: B019
```
abompard: Are you sure that this can't leak memory? It does not seem to be a singleton this time.

nphilipp: It’s the metaclass used for `ArtifactType`, so effectively a singleton (`cls` will always be `ArtifactType`). I’m not super happy with the implementation, should I make this a separate class method of `ArtifactType` instead and use it explicitly?

abompard: Yeah, maybe something like `has_name` and `has_value` would also make it clearer what we're testing here. Of course you only need the latter so no need to implement the other one ;-)

nphilipp: Heh, something like `has_name()` already exists even though it might not be well-known, e.g.: `"packages" in ArtifactType.__members__`

abompard (approved):

PR #835, opened 2023-03-31
Straighten out API for querying artifacts

### LINKED ISSUE #779: Clean up backend API
# Story

As a user of the FMN backend API such as a frontend developer,
I want that it sticks to REST conventions closely,
so I don’t run into surprises.

# Acceptance Criteria

- [x] The FMN backend API follows common conventions about REST APIs.

nphilipp:
This merges the `/artifacts/owned` endpoint into `/artifacts` and also shifts applying wildcards to name patterns from the backend to the frontend.

Fixes: #779

### DIFF
--- a/fmn/api/handlers/misc.py
+++ b/fmn/api/handlers/misc.py
@@ -1,5 +1,6 @@
 import logging
 from importlib import metadata
+from itertools import chain
 
 from fastapi import APIRouter, Depends, HTTPException, Query
 from sqlalchemy.ext.asyncio import AsyncSession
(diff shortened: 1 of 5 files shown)


fmn/api/handlers/misc.py
```diff
@@ -37,39 +38,50 @@ def get_applications():
...
+            if ArtifactType.has_value(project["namespace"])
+        )
 
     for group in groups:
-        for project in await distgit_proxy.get_group_projects(name=group):
-            artifacts[project["fullname"]] = {"type": project["namespace"], "name": project["name"]}
+        artifacts.update(
+            (project["fullname"], {"type": project["namespace"], "name": project["name"]})
+            for project in await distgit_proxy.get_group_projects(name=group)
+            if ArtifactType.has_value(project["namespace"])
```
abompard, line 72: It feels like these 3 blocks have a lot in common. Would there be ways to share some code?

fmn/api/handlers/misc.py
```diff
@@ -37,39 +38,50 @@ def get_applications():
...
+        )
 
     return sorted(artifacts.values(), key=lambda a: (a["name"], a["type"]))
```
abompard: It looks like you are using a dict for de-duplication, because the values are dicts, and therefore not hashable. How about something like this:
```python
artifacts = set()
for ...:
    artifacts.add(
        (project["name"], project["namespace"])
        for project in await distgit_proxy....
        if ArtifactType.has_value(project["namespace"])
    )
return [{"name": a[0], "type": a[1]} for a in sorted(artifacts)]
```

fmn/api/handlers/misc.py
nphilipp: Good call!

fmn/api/handlers/misc.py
nphilipp: This together with the suggestion above should make the handler pretty slim, I like it.

abompard (approved):

PR #836, opened 2023-03-31
Consistently show artifact type in the frontend

nphilipp:
Previously, newly added artifacts only were listed with their names but when editing, existing artifacts were shown with their types.

In this example, `rpms/0ad` was an existing artifact tracked by the rule, `apron` was added:

image


abompard (approved):
Thanks. You are now a Javascript programmer.

nphilipp:
> Thanks. You are now a Javascript programmer.

By the Power of Cargo Cult!

PR #847, opened 2023-04-12
Add test for adminlink component

ryanlerch:


frontend/src/components/AdminLink.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    loginUser(userStore);
+    const { queryAllByText } = render(AdminLink);
+    expect(queryAllByText("Admin")).toStrictEqual([]);
+  });
+
+  it("renders the admin link when not logged in as an admin", () => {
+    const userStore = useUserStore();
+    loginAdmin(userStore);
+    const { getByText } = render(AdminLink);
+    getByText("Admin");
```
abompard: I would do something like  `expect(getByText("Admin")).toBeInTheDocument()`, it seems clearer, even though your line will correctly raise an exception if not found.

frontend/src/components/TestUtils.ts
```diff
@@ -0,0 +1,48 @@
+// SPDX-FileCopyrightText: Contributors to the Fedora Project
```
abompard: This file should be named `testutils.ts` or `test-utils.ts`, the CamelCase is reserved to components (as a convention).

abompard (changes requested):
Looks good! Just very minor comments.

PR #855, opened 2023-04-12
Work with case-normalized application names

### LINKED ISSUE #854: Work with case-normalized application names
# Story

As a user of FMN,
I want that application names in messaging schemas are treated case-insensitively,
so that I don’t have to choose between `Copr` and `copr`.

# Acceptance Criteria

- [x] For the underlying logic of FMN, the case of application names does not matter
  - [x] API: only present one case-variant to frontend
  - [x] Filters: normalize app names to lowercase
  - [x] Filters: compare app names case-insensitively

# Background

Version 0.9 of `copr-messaging` explicitly defines `app_name` to `Copr`, the schema defines old versions of messages without (where it defaults to `copr`). FMN should be robust against such simple inconsistencies in case.

nphilipp:
Fixes: #854

### DIFF
--- a/fmn/api/handlers/misc.py
+++ b/fmn/api/handlers/misc.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: MIT
 
 import logging
+from functools import cache
 from importlib import metadata
 from itertools import chain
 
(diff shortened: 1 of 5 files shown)


nphilipp (commented):
Looks good to me, but I would say that!

PR #856, opened 2023-04-12
Better handling of sender errors

abompard:
Quite a few commits in this PR, I hope they are clear enough.

Now, when a protocol handler errors out, the sender component should shutdown (and get restarted by OpenShift)

### DIFF
--- a/fmn/sender/cli.py
+++ b/fmn/sender/cli.py
@@ -3,7 +3,6 @@
 # SPDX-License-Identifier: MIT
 
 import asyncio
-from functools import partial
 
 import click
 
@@ -14,15 +13,14 @@ from .handler import HandlerError
 HANDLER_CONNECT_TIMEOUT = 60
 
 
-def shutdown(result, consumer):
+async def _shutdown(result, consumer):
     if asyncio.isfuture(result):
-        result = result.result()
+        result = await result
     click.echo(f"Shutting down: {result}")
-    loop = asyncio.get_event_loop()
-    if loop.is_running():
-        loop.create_task(consumer.stop())
-    else:
-        loop.run_until_complete(consumer.stop())
+    await consumer.stop()
+
+
+background_tasks = set()
 
 
 async def _main(handler, consumer):
@@ -32,8 +30,14 @@ async def _main(handler, consumer):
         raise HandlerError(f"the handler could not connect in {HANDLER_CONNECT_TIMEOUT}s") from e
     except HandlerError as e:
         raise HandlerError(f"the handler could not connect: {e}") from e
+
     # Shutdown in case of unexpected disconnections
-    handler.closed.add_done_callback(partial(shutdown, consumer=consumer))
+    shutdown_on_closed = asyncio.create_task(_shutdown(handler.closed, consumer))
+    # Trick Python into creating a strong reference
+    # https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task
+    background_tasks.add(shutdown_on_closed)
+    shutdown_on_closed.add_done_callback(background_tasks.discard)
+
     await consumer.connect()
     await consumer.start()
 
@@ -55,5 +59,5 @@ def main(config_path):
     try:
         loop.run_until_complete(_main(handler, consumer))
     except Exception as e:
-        shutdown("exception caught", consumer)
+        loop.run_until_complete(_shutdown("exception caught", consumer))
         raise click.ClickException(e) from e
--- a/fmn/sender/cli.py
+++ b/fmn/sender/cli.py
@@ -28,8 +28,6 @@ async def _main(handler, consumer):
         await asyncio.wait_for(handler.setup(), timeout=HANDLER_CONNECT_TIMEOUT)
     except asyncio.exceptions.TimeoutError as e:
         raise HandlerError(f"the handler could not connect in {HANDLER_CONNECT_TIMEOUT}s") from e
-    except HandlerError as e:
-        raise HandlerError(f"the handler could not connect: {e}") from e
 
     # Shutdown in case of unexpected disconnections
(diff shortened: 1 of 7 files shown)


fmn/sender/cli.py
```diff
@@ -3,15 +3,37 @@
...
 from .config import get_config, get_handler, setup_logging
 from .consumer import Consumer
+from .handler import HandlerError
+
+HANDLER_CONNECT_TIMEOUT = 60
+
+
+def shutdown(result, consumer):
+    if asyncio.isfuture(result):
+        result = result.result()
```
nphilipp: Is it guaranteed that the `result` Future is done here?

```
…
asyncio.Future = class Future(builtins.object)
 |  asyncio.Future(*, loop=None)
 |  
 |  This class is *almost* compatible with concurrent.futures.Future.
 |  
 |  Differences:
 |  
 |  - result() and exception() do not take a timeout argument and
 |    raise an exception when the future isn't done yet.
…
```

Maybe let’s do this instead:

```
…
    if asyncio.isfuture(result):
        result = await result
…
```

fmn/sender/cli.py
```diff
@@ -3,15 +3,37 @@
...
 
 
 async def _main(handler, consumer):
-    await handler.setup()
+    try:
+        await asyncio.wait_for(handler.setup(), timeout=HANDLER_CONNECT_TIMEOUT)
+    except asyncio.exceptions.TimeoutError as e:
+        raise HandlerError(f"the handler could not connect in {HANDLER_CONNECT_TIMEOUT}s") from e
+    except HandlerError as e:
+        raise HandlerError(f"the handler could not connect: {e}") from e
```
nphilipp: Why wrap the `TimeoutError` twice in a `HandlerError`?

fmn/sender/handler.py
```diff
@@ -2,10 +2,14 @@
...
 # SPDX-License-Identifier: MIT
 
+import asyncio
+
 
 class Handler:
     def __init__(self, config):
         self._config = config
+        # Triggered when there is an error and the app should stop
+        self.closed = asyncio.get_event_loop().create_future()
```
nphilipp: Hmm, this looks like it’s only in place to ensure that any handler supports the `closed` Future, but it’s never really used except in `IRCHandler` where it dispatches to the aggregated `IRCClient` object and causes the `Handler.closed` Future to be thrown away. How about late initialization using a (cached) property? 

```
…
from functools import cached_property
…
class Handler:
    def __init__(self, config):
        self._config = config
        
   @cached_property
   def closed(self):
        """Default `closed` Future, can be overridden in child classes."""
        return asyncio.get_event_loop().create_future()
…
```

… and …

```
…
class IRCHandler(Handler):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._client = IRCClient()

    @property
    def closed(self):
        return self._client.closed
…
```

tests/conftest.py
```diff
@@ -80,8 +81,9 @@ def cashews_cache(monkeypatch: pytest.MonkeyPatch, request: pytest.FixtureReques
     with ctxmgr:
         yield
 
-    cache.clear()
-    cache.close()
+    with suppress(asyncio.exceptions.CancelledError):
+        await cache.clear()
+        await cache.close()
```
nphilipp, line 88: 👍

tests/sender/test_irc.py
```diff
@@ -34,15 +54,48 @@ async def _send_welcome():
...
+    _send_event(handler, transport, error_type, "server", "user", ["dummy error"])
+    await handler.closed
+
+
+async def test_irc_disconnect_expected(handler):
+    _send_event(handler, transport, "900", "server", "user")
+    await handler.setup()
+    await handler.stop()
+    # We didn't call the shutdown handler
+    assert handler.closed.done() is False
```
nphilipp:
```suggestion
    assert not handler.closed.done()
```

tests/sender/test_irc.py
```diff
@@ -34,15 +54,48 @@ async def _send_welcome():
...
+
+
+async def test_irc_disconnect_not_connected(handler):
+    handler._client.connected = False
+    await handler.stop()
+    # We didn't call the shutdown handler
+    assert handler.closed.done() is False
```
nphilipp:
```suggestion
    assert not handler.closed.done()
```

nphilipp (changes requested):

fmn/sender/cli.py
abompard: Ah, but that would make the function async, which it can't be if used in `add_done_callback`. I'll do it a bit differently.

fmn/sender/cli.py
abompard: I'm not sure I understand. The exception raised in the `TimeoutError` handler should not be caught by the second exception handler, no? Or is it? 
I can move the `HandlerError` exception handler first if it makes things clearer.

fmn/sender/handler.py
abompard: Yep, I'll do that!

fmn/sender/cli.py
nphilipp: > Ah, but that would make the function async, …

Oops, totally missed that! 🙈

fmn/sender/cli.py
nphilipp: > I'm not sure I understand. The exception raised in the `TimeoutError` handler should not be caught by the second exception handler, no? Or is it?

Hmm, I did some testing in the REPL and it looked like it was, but I did it again, now it doesn’t… Not sure what I did wrong before.

Still, I don’t understand why `except HandlerError` rewraps the exception in another `HandlerError`.

abompard: Yeah, that's not really necessary, I'll fix that.

nphilipp (approved):
LGTM!

PR #920, opened 2023-06-07
Add a way to extend the email message body

### LINKED ISSUE #916: Missing commit diff in "[Pagure] <fas> pushed 1 commits on rpms/<pkg> (branch: main)" mails
I am sorry, but the new FMN mails for commits to src.fedoraproject.org lost somehow their value for me; example:

```
Date: Fri,  2 Jun 2023 19:27:17 +0000 (UTC)
From: Fedora Notifications <notifications@fedoraproject.org>
To: <me>
Subject: [Pagure] <fas> pushed 1 commits on rpms/<pkg> (branch: main)

New commit: 1 commits
By: <fas>
https://src.fedoraproject.org/rpms/<pkg>/tree/main
```

In difference, the old FMN mails contained a diff of the actual Git commit (including the commit hash)…which I kindly would prefer to receive again :)

abompard:
This will allow schemas to define a `patch_url` attribute that will be retrieved when the email notification is generated.

Fixes: #916

### DIFF
--- a/fmn/database/model/destination.py
+++ b/fmn/database/model/destination.py
@@ -2,8 +2,10 @@
 #
 # SPDX-License-Identifier: MIT
 
+import logging
 from typing import TYPE_CHECKING
 
+from httpx import AsyncClient, HTTPStatusError
 from sqlalchemy import Column, ForeignKey, Integer, String, UnicodeText
 from sqlalchemy.orm import relationship
 
@@ -14,6 +16,9 @@ if TYPE_CHECKING:
     from fedora_messaging.message import Message
 
     from ...rules.notification import Notification
+
+
+log = logging.getLogger(__name__)
 
 
 class Destination(Base):
@@ -29,16 +34,20 @@ class Destination(Base):
     protocol = Column(String(length=255), nullable=False)
     address = Column(UnicodeText, nullable=False)
 
-    def generate(self, message: "Message") -> "Notification.content":
+    async def generate(self, message: "Message") -> "Notification.content":
         app_name = f"[{message.app_name}] " if message.app_name else ""
         url = message.url if message.url else ""
         if self.protocol == "email":
+            body = f"{str(message)}\n{url}"
+            extra = await get_extra(message)
+            if extra:
+                body = f"{body}\n{extra}"
             return {
                 "headers": {
                     "To": self.address,
                     "Subject": f"{app_name}{message.summary}",
                 },
-                "body": f"{str(message)}\n{url}",
+                "body": body,
             }
         elif self.protocol == "irc":
             return {"to": self.address, "message": f"{app_name}{message.summary} {url}"}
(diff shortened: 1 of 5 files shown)


fmn/database/model/destination.py
```diff
@@ -29,20 +34,34 @@ class Destination(Base):
...
             return {"to": self.address, "message": f"{app_name}{message.summary} {url}"}
         elif self.protocol == "matrix":
             return {"to": self.address, "message": f"{app_name}{message.summary} {url}"}
         else:
             raise ValueError(f"Unknown destination protocol: {self.protocol}")
+
+
+async def get_extra(message: "Message"):
+    http_client = AsyncClient(timeout=10)
+    if getattr(message, "patch_url", None) is not None:
```
nphilipp, line 62:
```suggestion
    if hasattr(message, "patch_url"):
```

fmn/database/model/destination.py
```diff
@@ -29,20 +34,34 @@ class Destination(Base):
...
         url = message.url if message.url else ""
         if self.protocol == "email":
+            extra = await get_extra(message)
             return {
                 "headers": {
                     "To": self.address,
                     "Subject": f"{app_name}{message.summary}",
                 },
-                "body": f"{str(message)}\n{url}",
+                "body": f"{str(message)}\n{url}\n{extra}",
```
nphilipp: If there is no `extra`, this leaves a stray trailing newline. How about:

```
            extra = await get_extra(message)
            if extra:
                body = f"{str(message)}\n{url}\n{extra}"
            else:
                body = f"{str(message)}\n{url}"
            return {
                "headers": {
                    "To": self.address,
                    "Subject": f"{app_name}{message.summary}",
                },
                "body": body,
```

nphilipp (changes requested):

fmn/database/model/destination.py
abompard: Well, not only, I also want to abort if `message.patch_url is None`

fmn/database/model/destination.py
abompard: Yep, will merge

fmn/database/model/destination.py
nphilipp: And if `message.patch_url` is an empty string? Then it could be `if getattr(message, "patch_url):`.

abompard: Hmm, I don't see why it would be an empty string, that seems wrong. The semantics I thought were:
- schema does not support patch_url: None through getattr
- schema supports patch_url but can't determine what it is: schema returns None
- schema supports patch_url and can determine it: it's a real URL

I don't see why a schema would return an empty string.

nphilipp: Ah, understood.

fmn/database/model/destination.py
nphilipp: 👍

nphilipp (approved):
LGTM!

PR #1226, opened 2024-10-25
Add a `Date` header to outgoing email

### LINKED ISSUE #1145: Delayed notifications have misleading timestamps.
Like its predecessor, the new notification system can send emails several days after the event happened. Those emails seem to be dated when the notification was converted to email form, and contain no information about when the event happened. The misleading timestamp makes it difficult to figure out whether something actually happened, or whether the notification pertains to an old event that I already know about. (Notifications about Git pushes contain a timestamp in the message body, but I think that's Git's "commit date" or "author date", not the time when the commit was pushed.)

The Date header field in the email should be set to the time when the event happened, not the time when the notification was converted from Fedora messaging form to email form.

According to RFC 5321 the gateway that converts the notification to email form must add a Received field. That's the field that shall contain the time when the conversion happened:
https://datatracker.ietf.org/doc/html/rfc5321#section-3.7.2
https://datatracker.ietf.org/doc/html/rfc5321#section-4.4

abompard:
Fixes: #1145

### DIFF
--- a/changelog.d/1145.fixed
+++ b/changelog.d/1145.fixed
@@ -1,0 +1,1 @@
+Add a `Date` header to outgoing email
--- a/fmn/core/amqp.py
+++ b/fmn/core/amqp.py
@@ -2,10 +2,15 @@
 #
 # SPDX-License-Identifier: MIT
 
+import logging
 import ssl
+from datetime import datetime, timezone
 
 from aio_pika.abc import SSLOptions
 from aio_pika.connection import URL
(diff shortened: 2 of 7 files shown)


abompard:
2 necessary changes according to https://github.com/fedora-infra/fmn/issues/1145#issuecomment-2438292817 :

- Set the Date to the message's `sent-at` header value
- Add a Received header because FMN acts as a "gateway"

PR #1375, opened 2025-08-25
UI: Guard against data loss in rule creation (#897)

### LINKED ISSUE #897: It's a bit too easy to miss the final commit step, in creating a rule.
When going through the steps of creating a new rule:

1. Click "Add a new rule"
2. "Choose a Tracking Rule" and set an optional "Rule Title"
3. Click the just-appeared "+Add Destination" button
4. Fill out the destination form and click "Add Destination" (again, but a different one)

...having done that, it's a little bit too easy to miss that you haven't _actually_ created **anything** unless you hit the "Create Rule" button in the upper-right corner, which is what finally does the actual rule-creating.

If the user mistakenly thinks they've already set everything they need for their rule, and instead clicks the "Notifications" logo instead, the interface will (without so much as an "Are you sure?" prompt) happily wander away from the in-progress rule creation, discarding all of the just-configured parameters.

This perhaps somewhat ties in to #877, since having a visible Cancel button would hopefully make it more clear that there's an in-progress activity yet to be completed.

sgollor:
# What does this change do?
- Renames "Create Rule" button to 'Save Rule" for clarity
- Moves the action buttons from the top-right to the bottom-right of the form.
- Warning messages: When the user tries to leave the page (mid-creation) without saving the new rule, and when they click the cancel button. 

Fixes: #897 

# How should we test your change?
1. Start the FMN frontend (pnpm dev).
2.  Go to Create a new Rule (/dev/new-rule).
3. Type a rule title and select a tracking rule.
4. Try each of these scenarios:
    - Click Cancel; should prompt once to discard changes.
    - Navigate to another route inside the app; should prompt once.
    - Refresh or close the tab; browser shows a warning message.
    
UI Screenshot
image

-

# Checklist for Submitter
- [x] Manual test plan described
- [x] Screenshots attached for UI changes
- [x] No database changes required

### DIFF
--- a/frontend/src/components/NewRuleForm.vue
+++ b/frontend/src/components/NewRuleForm.vue
@@ -27,11 +27,10 @@ const trackingRuleName = ref("");
 
 const generationRulesCount = ref(0);
 
-// Dirty state to track unsaved changes
-const isDirty = ref(false)
-const markDirty = () => {
-  isDirty.value = true
-}
+const newRuleForm = ref<{ node: FormKitNode } | null>(null);
+const isDirty = computed(
+  () => newRuleForm.value?.node.context?.state.dirty === true,
+);
 
 const formReady = computed(
   () => trackingRuleName.value !== "" && generationRulesCount.value > 0,
@@ -48,7 +47,6 @@ const handleSubmit = async (
   try {
     await mutateAsync(data as Rule);
     // Success!
-    isDirty.value = false;
     toastStore.addToast({
       color: "success",
       title: "Rule created",
@@ -65,28 +63,22 @@ const handleSubmit = async (
   }
 };
 
-// Mark form to have unsaved changes on input
 const handleTrackingRuleSelected = (name: string) => {
   trackingRuleName.value = name;
-  markDirty();
 };
 const handleGenerationRulesChanged = (rules: GenerationRule[]) => {
   generationRulesCount.value = rules.length;
-  markDirty();
 };
 
-// Add a a guarded cancel button to avoid losing unsaved changes
-const onCancel = async () => {
-  if (!isDirty.value || window.confirm("Discard the unsaved changes?")) {
-    isDirty.value = false;
-    router.push("/");
-  }
+const onCancel = () => {
+  router.push("/");
 };
 
 // Warning on tab close or hard navigation away
 const beforeUnloadHandler = (e: BeforeUnloadEvent) => {
   if (!isDirty.value) return;
   e.preventDefault();
+  e.returnValue = "";
 };
 onMounted(() => {
   window.addEventListener("beforeunload", beforeUnloadHandler);
@@ -101,14 +93,14 @@ onBeforeRouteLeave((_to, _from, next) => {
     next();
     return;
   }
-  if (window.confirm("You have in-app unsaved changes. Discard them?")) next();
+  if (window.confirm("You have not saved the new rule yet. Discard new rule?")) next();
   else next(false)
 });
 
 </script>
 
 <template>
-  <FormKit type="form" id="new-rule" @submit="handleSubmit" :actions="false" @input="markDirty">
+  <FormKit type="form" id="new-rule" @submit="handleSubmit" :actions="false" ref="newRuleForm">
     <!-- Track if the user has made changes in the form-->  
 
     <CRow class="mb-2 align-items-center">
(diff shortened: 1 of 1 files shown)


sgollor:
Greetings, this PR addresses #897. Can we get a reviewer with write access and or a maintainer to approve so we can close? @abompard @relrod

frontend/src/components/NewRuleForm.vue
```diff
@@ -26,6 +26,13 @@ const { mutateAsync } = useAddRuleMutation();
 const trackingRuleName = ref("");
 
 const generationRulesCount = ref(0);
+
+// Dirty state to track unsaved changes
+const isDirty = ref(false)
+const markDirty = () => {
+  isDirty.value = true
+}
+
```
abompard: You don't actually need to track the dirtiness manually, FormKit does it automatically. You can use its state in the `context` property. For example:
```suggestion
const newRuleForm = ref<{ node: FormKitNode } | null>(null);
const isDirty = computed(
  () => newRuleForm.value?.node.context?.state.dirty === true,
);
```

You can then remove `markDirty` everywhere.

frontend/src/components/NewRuleForm.vue
```diff
@@ -57,37 +65,56 @@ const handleSubmit = async (
...
+  window.removeEventListener("beforeunload", beforeUnloadHandler);
+});
+
+// Warn on in-app navigation away/route changes
+onBeforeRouteLeave((_to, _from, next) => {
+  if (!isDirty.value) {
+    next();
+    return;
+  }
+  if (window.confirm("You have in-app unsaved changes. Discard them?")) next();
```
abompard: This message is a little generic, we can do something more meaningful. What about `"You have not saved the new rule yet. Abort its creation?"` ?

abompard (changes requested):
Thanks for your work! I have a couple suggestions.

ferdnyc:
> * Warning messages: When the user tries to leave the page (mid-creation) without saving the new rule, and when they click the cancel button.

I wonder if a warning when clicking Cancel might be a bit much? It's not like the user doesn't know that, when they click Cancel, their work is going to be discarded. It's WHY they clicked Cancel.

A warning on what appears to be _accidental_ navigation, I'm very much in favor of.

frontend/src/components/NewRuleForm.vue
ferdnyc: The button labels in the resulting dialog are "Cancel" and "OK" (at least in Chrome, in English) — I'm worried that "Abort its creation?" might be a confusing prompt. ("So, I should click Cancel if I want to stop aborting my rule creation?" :face_with_spiral_eyes: )

Something like "Discard changes?" or "Discard new rule?" does feel more obvious. Though I can' t really give a good justification _why_, because it has the exact same double-negative issue, really. It just... flows more naturally?

abompard: That's a very good point, I'm not a native speaker so I'll defer to you. I was mostly thinking about the start of the function, mentioning that it's about the new rule.

sgollor:
Thanks for the feedback. I've made some updates to use FormKit to track changes automatically and also adjusted the confirmation messages to be clearer. @abompard @ferdnyc 

Kindly review and merge/close if it is okay.

frontend/src/components/NewRuleForm.vue
```diff
@@ -63,31 +69,44 @@ const handleTrackingRuleSelected = (name: string) => {
...
+  window.removeEventListener("beforeunload", beforeUnloadHandler);
+});
+
+// Warn on in-app navigation away/route changes
+onBeforeRouteLeave((_to, _from, next) => {
+  if (!isDirty.value) {
+    next();
+    return;
+  }
+  if (window.confirm("Discard new rule?")) next();
```
ferdnyc: Oh, sorry, that's my fault for being unclear. I wasn't suggesting that the _entire_ message be just that question, I was merely suggesting to use it as the _prompt_ at the end of @abompard's suggested message. So, more like...

```suggestion
  if (window.confirm("You have not saved the new rule yet. Discard new rule?")) next();
```

or

```suggestion
  if (window.confirm("You have not saved the new rule yet. Discard changes?")) next();
```

sgollor: Message corrected @ferdnyc

ferdnyc: LGTM, but I'm nobody here, my review carries no weight (procedurally). @abompard or another committer will have to weigh in.

sgollor: @abompard Over to you. Let's go! 💪

abompard (approved):
Yep, looks good, thanks for your contribution!

sgollor:
Thank you very much for your time and feedbacks! 🙏


________________________________

PR #1378, opened 2025-09-07
UI: clarify blank means All in Applications (#876)

### LINKED ISSUE #876: [RFE] Provide "all applications" to rules creation
On creating a new rule (for example for "My Events" or "My artifacts") I would like to select "all applications", so in future when new apps are added I don't have to remember to update my rule.

sgollor:
# What does this change do?
This pull request adds helper text to clarify that leaving the Applications field blank/unselected means the rule applies to all applications (including new ones added later).

It also adds a tooltip icon next to the label with the text: “Leaving this unselected means your rule applies to every application.”

Fixes: #876

# How should we test your change?

1. Run the frontend locally (npm run dev).
2. Open the New Rule form.
3. In the Applications field:
    - Hover over ⓘ to see the tooltip.
4. Confirm helper text is visible under the field.

-

image

### DIFF
--- a/frontend/src/components/rule-edit/generation-rule/BaseFilterApplication.vue
+++ b/frontend/src/components/rule-edit/generation-rule/BaseFilterApplication.vue
@@ -11,6 +11,7 @@ const props = defineProps<{
   name: string;
   label: string;
   placeholder?: string;
+  help?: string;
 }>();
 
 const url = "/api/v1/applications";
@@ -33,5 +34,6 @@ const getApplications = async () => {
     :msOptions="getApplications"
     searchable
     :close-on-select="false"
+    :help="props.help"
   />
 </template>
--- a/frontend/src/components/rule-edit/generation-rule/FilterApplication.vue
+++ b/frontend/src/components/rule-edit/generation-rule/FilterApplication.vue
@@ -13,5 +13,6 @@ import BaseFilterApplication from "./BaseFilterApplication.vue";
     name="applications"
     label="Applications"
     placeholder="Choose applications"
+    help="If you leave this field empty, the rule applies to all applications (including new ones added later)"
   />
 </template>
--- a/frontend/src/forms/index.ts
+++ b/frontend/src/forms/index.ts
@@ -53,6 +53,12 @@ export const config = defaultConfig({
         input: "form-check-input",
         help: "text-muted",
       },
+      multiselect: {
+        help: "text-muted",
+      },
+      multiselectasyncdefault: {
+        help: "text-muted",
+      },
     }),
   },
 });


frontend/src/components/rule-edit/generation-rule/BaseFilterApplication.vue
```diff
@@ -24,7 +25,16 @@ const getApplications = async () => {
...
+  <!-- Label with tooltip -->
+  <p class="mb-0 fw-bold">{{ props.label }}
+    <span
+      aria-label="help"
+      class="ml-1 cursor-help text-muted"
+      title="Leaving this unselected means your rule applies to every application"
+    >ⓘ <!-- tooltip icon -->
+  </span>
+  :</p>
+
```
abompard: That won't work because the base component is used for the `FilterApplication` component *and* the `FilterExcludedApplication` component, and we don't want that tooltip for both.

frontend/src/components/rule-edit/generation-rule/FilterApplication.vue
```diff
@@ -13,5 +13,6 @@ import BaseFilterApplication from "./BaseFilterApplication.vue";
     name="applications"
     label="Applications"
     placeholder="Choose applications"
+    help="If you leave this field empty, the rule applies to all applications (including new ones added later)"
```
abompard, line 16: The text does not appear in gray as it does for the topic pattern widget help message.
I think you have to edit `frontend/src/forms/index.ts` and add these to the FormKit classes:

```js
multiselect: {
  help: "text-muted",
},
multiselectasyncdefault: {
  help: "text-muted",
},
```

abompard (changes requested):
Thanks!

sgollor:
@abompard Thank you very much for the feedback. I have made some updates.

image

frontend/src/components/rule-edit/generation-rule/FilterApplication.vue
```diff
@@ -13,5 +13,18 @@ import BaseFilterApplication from "./BaseFilterApplication.vue";
...
+      <span>
+        Applications
+        <span
+          aria-label="help"
+          class="ml-1 cursor-help text-muted"
+          title="Leaving this unselected means your rule applies to every application"
+        >ⓘ <!-- tooltip icon -->
+        </span>
+      </span>
+    </template>
```
abompard: I'm not sure it's useful to have a tooltip on the label: the message is redundant with the help text, and tooltips aren't visible on mobile (because there's no hover).
Please remove the slot.

frontend/src/forms/index.ts
```diff
@@ -53,6 +53,12 @@ export const config = defaultConfig({
         input: "form-check-input",
         help: "text-muted",
       },
+       multiselect: {
```
abompard: Extra whitespace here.

abompard (changes requested):

frontend/src/forms/index.ts
sgollor: Extra whitespace removed

frontend/src/components/rule-edit/generation-rule/FilterApplication.vue
sgollor: Slot removed

abompard (approved):
Looks good, thanks for your contribution!

sgollor:
Thank you very much for your valuable feedback and guidance. Let's go! 🙏💪

